PR #4, opened 2018-12-31
Use @babel/runtime helpers

Andarist:
Looking through emotion's packages:
- `@emotion/core` is using `inheritsLoose`
- `emotion-theming` is using `extends` (shouldnt it use object-assign though? 🤔)

So not much - but still IMHO a general optimization that should be done.


src/rollup-plugins/rewrite-cjs-runtime-helpers.js
```diff
@@ -0,0 +1,48 @@
+import * as babel from "@babel/core";
+
+const helperPath = /^@babel\/runtime\/helpers\/esm\/(\w+)$/;
+
+export default function rewriteCjsRuntimeHelpers() {
```
mitchellhamilton: Why doesn't babel do this by default/have package.jsons that redirect?

Andarist: `package.json` redirects are not a popular technique and unfortunately they were not implemented for those, I've started a PR for this some time ago - https://github.com/babel/babel/pull/8885 but didnt get it merged.

There is a problem though - `babel` inserts requires calls to those helpers without interop helper and they just do `module.exports =`, which means their cjs shape is different than esm and unfortunately webpack@4+ doesnt try to interop such situation ( https://github.com/webpack/webpack/issues/7973 ).

So basically to make it work right now in babel we would have to introduce another set of files that would reexport those helpers as .default exports in cjs format and redirect to those, instead of regular cjs entries.

Kinda a mess, but it is what it is :/

mitchellhamilton: K, gonna merge this.

PR #93, opened 2019-09-09
Add typescript types support

torifat:
Hi,

I was trying use `preconstruct` in our `typescript` mono-repo but I found that it doesn't fix the `types` field. So, I added it. Let me know if you want me to change anything.

Cheers

### DIFF
--- a/__fixtures__/typescript/package.json
+++ b/__fixtures__/typescript/package.json
@@ -5,6 +5,7 @@
   "license": "MIT",
   "private": true,
   "module": "dist/typescript.esm.js",
+  "types": "dist/typescript.cjs.js.ts",
   "dependencies": {
     "@babel/core": "^7.4.3",
     "@babel/preset-typescript": "^7.3.3",
--- a/packages/preconstruct/src/__tests__/fix.js
+++ b/packages/preconstruct/src/__tests__/fix.js
@@ -221,3 +221,31 @@ Object {
     expect(promptInput).toBeCalledTimes(1);
   }
 );
+
+test("fix types", async () => {
+  let tmpPath = f.copy("typescript");
+
+  await modifyPkg(tmpPath, pkg => {
+    pkg.types = "bad.js";
+  });
+
+  await fix(tmpPath);
+
+  expect(await getPkg(tmpPath)).toMatchInlineSnapshot(`
+Object {
+  "dependencies": Object {
+    "@babel/core": "^7.4.3",
+    "@babel/preset-typescript": "^7.3.3",
+    "@types/node": "^12.7.1",
+    "typescript": "^3.4.5",
+  },
+  "license": "MIT",
+  "main": "dist/typescript.cjs.js",
+  "module": "dist/typescript.esm.js",
+  "name": "typescript",
+  "private": true,
+  "types": "dist/typescript.cjs.js.ts",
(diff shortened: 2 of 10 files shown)


torifat:
Do I need add a `changeset`?

torifat:
Thanks, it makes sense. I will update the PR soon :)

mrmartineau:
@mitchellhamilton it looks like this PR is dead. What exactly needs to happen (apart from fixing conflicts) for this to be in a merge-ready state? I'll happily take this over if possible. Let me know 👍

ericclemmons:
@mrmartineau You should be able to checkout this PR to your own fork `hub pr checkout 93` (https://github.com/github/hub), make changes, and submit a new PR.

That way @torifat gets credit for the contributions, and you can get it across the finish line. 🏁

mitchellhamilton:
That sounds great!! What needs to be done is what I mentioned in my last review of this PR

mrmartineau:
ok, I'll try to take a look this week. Cheers guys

mrmartineau:
@mitchellhamilton after looking into this a bit, I would appreciate some help implementing the fixes. Since this PR was opened, the codebase was converted to TypeScript, so those lines of code do not exist anymore:

> We should definitely fix for this case though. So if an entrypoint source file matches `/\.tsx?$/` and the entrypoint has the types or typings field, throw a fixable error in https://github.com/preconstruct/preconstruct/blob/master/packages/preconstruct/src/validate.js#L75-L118 and in https://github.com/preconstruct/preconstruct/blob/master/packages/preconstruct/src/fix.js#L10-L21 delete the fields.

These files are the new TS versions, which lines do you need amended, and how? 
- https://github.com/preconstruct/preconstruct/blob/master/packages/cli/src/validate.ts
- https://github.com/preconstruct/preconstruct/blob/master/packages/cli/src/fix.ts

Are you mostly happy with the rest of @torifat's changes?

mitchellhamilton:
https://github.com/preconstruct/preconstruct/blob/4b1f0b5e91e18bf2367708791c42ca31c0a57d2d/packages/cli/src/validate.ts#L74 needs to check that the `types` and `typings` field do not exist if the source file is a TypeScript(`/\.tsx?$/.test(entrypoint.source)`)

https://github.com/preconstruct/preconstruct/blob/4b1f0b5e91e18bf2367708791c42ca31c0a57d2d/packages/cli/src/fix.ts#L9 needs to remove the `types` and `typings` field do not exist if the source file is a TypeScript(`/\.tsx?$/.test(entrypoint.source)`)

https://github.com/preconstruct/preconstruct/blob/4b1f0b5e91e18bf2367708791c42ca31c0a57d2d/packages/cli/src/init.ts#L13 should also do the same as fix except here it needs to iterate over all of the entrypoints in there and prompt for whether or not Preconstruct should fix the field

Note that the handling of this field should be different to how other fields are handled because for other fields, we want to make sure it exists and Preconstruct has some rules around ensuring that if a field is on one entrypoint, it should be on all entrypoints but this does **not** apply to the types/typings field

The changes in this PR except for the tests maybe other small things which could be used with a couple small changes should not happen.

torifat:
Sorry, closing this since it became stale and I couldn't manage time to follow up :(

PR #103, opened 2019-10-04
adding documentation on entrypoints

Noviny:
`emoji-uid` was just a cute example package - should possibly not be this, or I should make `emoji-uid` work like this if you just run these commands.

Otherwise I think this gives good structure.

Thing-I-want-to-tackle-somewhere - explaining configs better. I don't know *how* because I just skipped the config index and just went for configs relating to what I wanted, but understanding that these are configs to go in different `package.json`s for different purposes feels like it needs to be referenced in each somehow neatly.

### DIFF
--- a/site/src/pages/config/entrypoints.mdx
+++ b/site/src/pages/config/entrypoints.mdx
@@ -6,7 +6,7 @@ Entrypoints are the lowest level configuration point and describe a set of bundl
 
 `string`
 
-`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified. This
+`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified.
 
 ### Default
 
--- a/site/src/pages/config/entrypoints.mdx
+++ b/site/src/pages/config/entrypoints.mdx
@@ -1,6 +1,6 @@
 # Entrypoints
 
-Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoing)
+Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoint)
 
 ## `source`
 
--- a/site/src/pages/config/entrypoints.mdx
+++ b/site/src/pages/config/entrypoints.mdx
@@ -1,12 +1,12 @@
 # Entrypoints
 
-Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint.
+Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoing)
 
 ## `source`
 
 `string`
 
-`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified.
+`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified. This
 
 ### Default
 
--- a/site/src/pages/guides/adding-a-second-entrypoint.mdx
+++ b/site/src/pages/guides/adding-a-second-entrypoint.mdx
@@ -75,7 +75,7 @@ The next thing we need to do is tell the new `awesome` entrypoint where to find 
 }
 ```
 
-This tells this entrypoint where to fine the source code it should bundle to here.
+This tells this entrypoint where to find the source code it should bundle to here.
 
 Finally, run
 
--- a/site/src/pages/guides/adding-a-second-entrypoint.mdx
+++ b/site/src/pages/guides/adding-a-second-entrypoint.mdx
@@ -1,0 +1,110 @@
+# Adding a second entrypoint
+
+When adding a second entrypoint, you are creating a second place where a package can be accessed from. Note that much of the time you do not need a second entrypoint. We have a guide on [when to use multiple entrypoints](/guides/when-should-i-use-multiple-entrypoints).
+
+So, let's assume we have a package called `emoji-uid` that gets us three random emojis, but we want to make a second function that only uses the best emojis. What we want to be able to do is write is:
+
+```js
+import getEmojiId from "emoji-uid";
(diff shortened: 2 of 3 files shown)


site/src/pages/config/entrypoints.mdx
```diff
@@ -1,12 +1,12 @@
...
 
-Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint.
+Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoing)
 
 ## `source`
 
 `string`
 
-`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified.
+`source` specifies the source file to use for a given entrypoint. It's resolved relative to the package.json where it's specified. This
```
mitchellhamilton: Unfinished sentence?

site/src/pages/config/entrypoints.mdx
```diff
@@ -1,12 +1,12 @@
 # Entrypoints
 
-Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint.
+Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoing)
```
mitchellhamilton:
```suggestion
Entrypoints are the lowest level configuration point and describe a set of bundles for a particular entrypoint. They are configured by the `package.json` in the folder of the entrypoint. We also have a guide on [adding a second entrypoint](/guides/adding-a-second-entrypoint)
```

site/src/pages/guides/adding-a-second-entrypoint.mdx
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+Let's assume we've written our new function and written it to `src/awesomenewFunction.js`
+
+The first bit of config we need to do is to add an `entrypoints` field to our preconstruct config. We should modify this in our package.json to read:
+
+```json
+  "preconstruct": {
+    "source": "src/index.js"
```
mitchellhamilton:
```suggestion
    "source": "src/index.js",
```

site/src/pages/guides/adding-a-second-entrypoint.mdx
```diff
@@ -0,0 +1,110 @@
...
+```
+
+This tells preconstruct to look for these two entrypoints. We have written out two, as the first is preconstruct's default entrypoint which we still want to include.
+
+Next, we will want to do to add the location the entrypoint resolves to. We are going to add a new directory with its own `package.json` that represents this. We can run:
+
+```bash
+mkdir awesome
+touch awesome/package.json
+```
```
mitchellhamilton: Not about this PR but btw, I think I have a way to make it so you don't need to create the directory but globs still work.

site/src/pages/guides/adding-a-second-entrypoint.mdx
```diff
@@ -0,0 +1,110 @@
...
+```json
+{
+  "main": "dist/emoji-uid.cjs.js",
+  "preconstruct": {
+    "source": "../src/awesomeNewFunction"
+  }
+}
+```
+
+This tells this entrypoint where to fine the source code it should bundle to here.
```
mitchellhamilton:
```suggestion
This tells this entrypoint where to find the source code it should bundle to here.
```

mitchellhamilton:
(Ignore that TS is failing, that's my fault)

site/src/pages/guides/adding-a-second-entrypoint.mdx
Noviny: 😆 how much mad science is it? How scared am I?

Super great though.

site/src/pages/guides/adding-a-second-entrypoint.mdx
Noviny: Mostly wanted to have an existing `preconstruct` field to add things to as a pointer to where to add things. Can make it something other than `src/index.js`.

site/src/pages/guides/adding-a-second-entrypoint.mdx
mitchellhamilton: It's not _that_ much mad science. The assumption is that a glob in an array of globs that isn't really a glob(it's just a path) should either match something or shouldn't be in the array of globs. If we use this assumption, we can go through the array of globs and check if it's a glob or not, if it's not, make sure that that the path/glob matches against the array of globs(just in case it's negated) and then check if the directory/package.json and if it doesn't prompt to create the entrypoint(and if they decline, remove the glob thing from the array of globs)

mitchellhamilton:
> emoji-uid was just a cute example package - should possibly not be this, or I should make emoji-uid work like this if you just run these commands.

I'm all for cute examples so I'm happy to leave it as this.

> Thing-I-want-to-tackle-somewhere - explaining configs better. I don't know how because I just skipped the config index and just went for configs relating to what I wanted, but understanding that these are configs to go in different package.jsons for different purposes feels like it needs to be referenced in each somehow neatly.

Agree that this needs to be better. This is a problem which I'm very unsure how to explain, I feel like it might be helpful to use a single package repo with one entrypoint and go "this thing is just like a monorepo but there's only one package and it only has one entrypoint, things with multiple package and entrypoints aren't really special"

site/src/pages/guides/adding-a-second-entrypoint.mdx
```diff
@@ -0,0 +1,110 @@
...
+```
+
+and our package.json looks like:
+
+```json
+{
+  "name": "emoji-uid",
+  "version": "1.0.2",
+  "description": "Get 🔥 UIDs to help make things 💯",
+  "main": "dist/emoji-uid/.cjs.js",
```
mitchellhamilton:
```suggestion
  "main": "dist/emoji-uid.cjs.js",
```

site/src/pages/guides/adding-a-second-entrypoint.mdx
mitchellhamilton: I don't mind - Happy to leave it how it is.

Noviny:
Apart from the config side-discussion is this good to merge? Feel we should move that somewhere else. Want me to make an issue or just leave us to keep brainstorming?

mitchellhamilton:
> Apart from the config side-discussion is this good to merge? Feel we should move that somewhere else. Want me to make an issue or just leave us to keep brainstorming?

An issue would be great

Issue #236, opened 2019-12-06
source file not found `typescript-declarations` PLUGIN_ERROR


mrmartineau:
Currently Preconstruct cannot build one of the entrypoints in my package (see the error message below). I have figured out what file is causing the issue, but not why the issue is being caused - does that make sense? Would it be possible to improve the feedback in the CLI so that debugging can be improved?

See below, the key part is on the 2nd line `source file not found` but it does not specify what source file..

```
🎁 info building bundles!
🎁 error my-design-system Error: This is an internal error, please open an issue if you see this: source file not found
🎁 error my-design-system     at searchDeps (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:881:34)
🎁 error my-design-system     at Object.getDeps (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:890:8)
🎁 error my-design-system     at Object.generateBundle (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:912:122) {
🎁 error my-design-system   code: 'PLUGIN_ERROR',
🎁 error my-design-system   plugin: 'typescript-declarations',
🎁 error my-design-system   hook: 'generateBundle'
🎁 error my-design-system }
```

Here is the source code that generated that error: https://github.com/preconstruct/preconstruct/blob/c4886ea432a839291437c3346ddc6253b462c918/packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts#L125-L155

By tweaking the source code in `node_modules` like below, I see that the problem file is displayed. So it would be great to generate an error message that included that path.

```diff
            throw new Error(
-              "This is an internal error, please open an issue if you see this: source file not found"
+             dep
            );
```

This now shows an error like so: 

```
🎁 info building bundles!
🎁 error my-design-system Error: /Users/my-design-system/src/themes/index.js
🎁 error my-design-system     at searchDeps (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:881:34)
🎁 error my-design-system     at Object.getDeps (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:890:8)
🎁 error my-design-system     at Object.generateBundle (/Users/my-design-system/node_modules/@preconstruct/cli/cli/dist/cli.cjs.prod.js:912:122) {
[...]

PR #239, opened 2019-12-10
feat: improve searchDeps error message content

mrmartineau:
This PR improves the error message given when there is a problem when TypeScript declarations could not be generated.

Ref issue: https://github.com/preconstruct/preconstruct/issues/236

### DIFF
--- a/packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
+++ b/packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
@@ -64,7 +64,8 @@ let getService = weakMemoize((typescript: Typescript) =>
 );
 
 export async function createDeclarationCreator(
-  dirname: string
+  dirname: string,
+  pkgName: string
 ): Promise<{
   getDeps: (entrypoints: Array<string>) => Set<string>;
   getDeclarationFile: (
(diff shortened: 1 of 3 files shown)


mrmartineau:
@mitchellhamilton I added a changeset, but I don't know if I did it correctly...

packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
```diff
@@ -126,8 +127,9 @@ export async function createDeclarationCreator(
         for (let dep of deps) {
           let sourceFile = program!.getSourceFile(dep);
           if (!sourceFile) {
-            throw new Error(
-              "This is an internal error, please open an issue if you see this: source file not found"
+            throw new FatalError(
+              `Could not generate type declarations because ${dep} does not exist or is not a TypeScript file`,
+              dep
```
mrmartineau: I wasn't sure what to have as the scope. Let me know if it should be something else

mitchellhamilton: It should be `pkg.name` so `createDeclarationCreator` should be changed to accept `Package`

mrmartineau: @mitchellhamilton FYI I passed in a new arg instead of all of `Package`

PR #258, opened 2020-02-17
Fixed issue with temporarily created files not being cleaned up

Andarist:
This could not get executed in full:
https://github.com/preconstruct/preconstruct/blob/d8dce6f0a3a9e144c14f6f0ff8c01bef18406fe0/packages/cli/src/validate-included-files.ts#L62-L70
because first thrown error was thrown up to the root-catch-handler and caused:
https://github.com/preconstruct/preconstruct/blob/1e37a5de036ca4f897c990457676c33a6ae287bc/packages/cli/src/cli.ts#L104

This, in turn, has terminated the program - in the middle of the cleanup.

### DIFF
--- a/packages/cli/src/project.ts
+++ b/packages/cli/src/project.ts
@@ -9,6 +9,16 @@ import { promptConfirm } from "./prompt";
 import { validateIncludedFiles } from "./validate-included-files";
 import { FatalError } from "./errors";
 
+const allSettled = (promises: Promise<any>[]) =>
+  Promise.all(
+    promises.map(promise =>
+      promise.then(
+        value => ({ status: "fulfilled" as const, value }),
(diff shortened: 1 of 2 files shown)


packages/cli/src/project.ts
```diff
@@ -9,6 +9,16 @@ import { promptConfirm } from "./prompt";
 import { validateIncludedFiles } from "./validate-included-files";
 import { FatalError } from "./errors";
 
+const allSettled = (promises: Promise<any>[]) =>
```
Andarist: based on https://tc39.es/proposal-promise-allSettled/#sec-promise.allsettled

mitchellhamilton:
Thanks!!

PR #261, opened 2020-02-19
Allow to build projects including usage of eval

Andarist:


packages/cli/src/build/rollup.ts
```diff
@@ -93,6 +93,7 @@ export let getRollupConfig = (
       }
       switch (warning.code) {
         case "EMPTY_BUNDLE":
+        case "EVAL":
```
Andarist: we could surface this as a warning, but I dont think this is really that much needed - usage of eval is so much discouraged in the community that it's quite obvious that you shouldnt use it (when not really really needed), without extra warnings/errors

PR #264, opened 2020-02-29
#263: stringify paths before generating dev flow/ts type system imports

### LINKED ISSUE #263: Typescript paths are not escaped properly in dist in dev mode
Expected:
`preconstruct dev` should properly set up typescript definitions for development

Actual:
Types are not wired up properly.

For example, in `card.cjs.js.ts`, we see the following generated code:
image

The slashes are not being properly escaped. This works:
image

The issue seems to be caused by `.\preconstruct\packages\cli\src\dev.ts`

```js
tsTemplate(
  hasDefaultExport,
  path
    .relative(path.dirname(cjsDistPath), entrypoint.source)
    .replace(/\.tsx?$/, "")
)
```

NateRadebaugh:
Resolves #263

Raw paths were being used in generated typescript/flow import code. However, this resulted in unescaped paths, which makes vscode sad.

image

Follows the pattern of other generated codes using `JSON.stringify` to escape paths, and wrap in `"` quotes.

image

### DIFF
--- a/.changeset/nine-sheep-cheer.md
+++ b/.changeset/nine-sheep-cheer.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Fix type import file paths when running `preconstruct dev` to point to properly escaped paths.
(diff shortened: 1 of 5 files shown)


NateRadebaugh:
@mitchellhamilton I had the same thought as soon as I sent out the PR, but went to bed. Updated PR to only update the utils

PR #266, opened 2020-03-01
#256: Normalize paths for declarations to fix TS builds on windows

### LINKED ISSUE #256: TypeScript projects break on Windows
When building a fresh repo using typescript on Windows. I've got path problems.

https://github.com/preconstruct/preconstruct/blob/e50769a843f827292bdf94dd52cb26be31f0fc1c/packages/cli/src/rollup-plugins/typescript-declarations/index.ts#L49-L56

where;

```plaintext
facadeModuleId == C:\Users\me\Sites\github\abc\packages\123\src\index.ts

srcFilenameToDtsFilenameMap == Map(1) {
  'C:/Users/me/Sites/github/abc/packages/123/src/index.ts' => 'C:/Users/me/Sites/github/abc/packages/123/src/index.d.ts'
}
```

as you can see, the indexes of the Map will never equal.

NateRadebaugh:
Resolves #256 

Unix vs windows have different slashes, `\\` vs `/`, which makes a `Map` fail to find TS declarations by path. Rather than refactoring the world to fix inconsistencies with `globby`, `path`, and `resolve`, we use `normalize-path` (which is already a dependency of a ton of other dependencies) before inserting into the `Map` and before calling `get`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2424,6 +2424,11 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-7.10.9.tgz#4343e3b009f8cf5e1ed685e36097b74b4101e880"
   integrity sha512-usSpgoUsRtO5xNV5YEPU8PPnHisFx8u0rokj1BPVn/hDF7zwUDzVLiuKZM38B7z8V2111Fj6kd4rGtQFUZpNOw==
 
+"@types/normalize-path@^3.0.0":
+  version "3.0.0"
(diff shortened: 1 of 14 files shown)


yarn.lock
```diff
@@ -7184,6 +7189,12 @@ global-prefix@^1.0.1:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/global-prefix/-/global-prefix-1.0.2.tgz#dbf743c6c14992593c655568cb66ed32c0122ebe"
   integrity sha1-2/dDxsFJklk8ZVVoy2btMsASLr4=
+  dependencies:
```
NateRadebaugh: Not sure why this got added to the `yarn.lock`, so I can revert this part of the change if you want.

PR #270, opened 2020-03-22
Update babel to support typescript 3.8 syntax

chenxsan:
Typescript 3.8 supports ["Type-Only Imports and Exports" feature](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/#type-only-imports-exports) which requires [babel 7.9.0](https://github.com/babel/babel/releases/tag/v7.9.0), but i don't know why the [`resolutions`](https://github.com/preconstruct/preconstruct/blob/master/package.json#L23) in package.json file failed to instruct yarn to install `@babel/parser^7.9.0` which means syntax like `import type { SomeType } from "./another-thing";` would fail when I use `preconstruct` to build typescript project:

image

This pull request update the `resolutions` field to use `@babel/parser^7.9.2`.


chenxsan:
Has nothing to do with the babel,

chenxsan:
Solved by specify `resolutions` in package.json for yarn 2:

```js
"resolutions": {
    "@babel/parser": "^7.9.2"
  }
```

PR #275, opened 2020-04-20
Experimental new entrypoints

mitchellhamilton:


Andarist:
> The important change here is that we are determining whether an entrypoint exists based on the source file rather than the output.

I'm a fan of this 😃 

> Will other which points to src/other.js, src/other/index.js or etc. work?

I'm not particularly a fan of index.js auto-resolving but I'm wondering if support for it wouldn't be a usability improvement here. Kinda similarly to globs. Imagine a component library (or a utility library like lodash) where each component should get its own entry - it would be troublesome to maintain manually the list of entries. However, you could maybe constrain supported patterns though, simple `components/*` is enough to cover this and for that to work you most likely would have to create entries based on `components/*/index.js`. 

> Which of these should we use?

As this will be auto-fixed/validated anyway I would vote for src-less list. It's easier to look at.

> TODOs

+ tests? :p
+ cleaning up removed entries?

packages/cli/src/package.ts
```diff
@@ -35,138 +37,237 @@ export class Package extends Item {
       this.name
     );
   }
-  static async create(directory: string): Promise<Package> {
+  static async create(directory: string, project: Project): Promise<Package> {
     let filePath = nodePath.join(directory, "package.json");
 
     let contents = await fs.readFile(filePath, "utf-8");
     let pkg = new Package(filePath, contents);
+    pkg.project = project;
```
Andarist, line 45: could `project` be passed to `Package` constructor to avoid assigning this after creation?

mitchellhamilton: Yeah, probably, was just trying to do this quickly to try it out in a project

mitchellhamilton:
> Imagine a component library (or a utility library like lodash) where each component should get its own entry - it would be troublesome to maintain manually the list of entries. However, you could maybe constrain supported patterns though, simple components/* is enough to cover this and for that to work you most likely would have to create entries based on components/*/index.js.

`components/*/index.js` would work and would create entries at `components/*`, does that satisfy this use case?

> cleaning up removed entries?

Interesting idea, how would you know what is a removed entry and what's a package.json that exists for some other reason?

Andarist:
> components/*/index.js would work and would create entries at components/*, does that satisfy this use case?

Yes, it should be documented and possibly validated with a helpful error if people try to use unsupported or confusing patterns.

> Interesting idea, how would you know what is a removed entry and what's a package.json that exists for some other reason?

Add an extra field to them - like `"preconstruct": true` or whatever. This would also allow for "preconstruct clean" command to be implemented - people might not want to commit all of those generated directories so such command could be helpful in that scenario as well.

mitchellhamilton:
> Yes, it should be documented and possibly validated with a helpful error if people try to use unsupported or confusing patterns.

💯 

> Add an extra field to them - like "preconstruct": true or whatever. This would also allow for "preconstruct clean" command to be implemented - people might not want to commit all of those generated directories so such command could be helpful in that scenario as well.

Yeah, that could work. I'm thinking it would probably be that entrypoint package.jsons(except for a package's package.json) existing in a package where they're not actually entrypoints would be invalid and we'd offer to remove them or otherwise exit with an error.

Andarist:
> Yeah, that could work. I'm thinking it would probably be that entrypoint package.jsons(except for a package's package.json) existing in a package where they're not actually entrypoints would be invalid and we'd offer to remove them or otherwise exit with an error.

Yeah, exactly - would be really nice to include an extra highlighted warning for cases when such invalid entry directory contains some extra files (something else than `package.json` & `dist` - which both are generated and handled through preconstruct)

mitchellhamilton:
I'm quite wary of adding warnings to Preconstruct because then you either always have to see a warning or we have to add a way to ignore warnings(errors are much better because you either fix them or throw). What would be the purpose behind warning on that case? When would you have something other than package.json and dist in there?

Andarist:
I've just misspoken - I've meant error.

> When would you have something other than package.json and dist in there?

It shouldn't really happen, just a super edge case I thought about - technically it's possible to add something there as preconstruct only "controls" package.json and dist directory, someone could add docs/tests/whatever there manually (not the best idea in the world, but possible)

Andarist:
btw - in the future it would make sense to use node's conditional exports for defining entry points, right?

mitchellhamilton:
maybe, not totally sure yet.

PR #278, opened 2020-04-20
fix: Force emit when generating declaration files

### LINKED ISSUE #271: getDeclarationFile function failing and triggering error
Builds are failing because getDeclarationFile can't get `output.outputFiles[0].name` 

one solution that I thought about was changing the function to have a condition to check if there is an `output.outputFiles[0]` object

```
getDeclarationFile: async filename => {
    if (filename.endsWith(".d.ts")) {
        return {
            name: filename.replace(dirname, path.join(dirname, "dist", "declarations")),
            content: await fs.readFile(filename, "utf8")
        };
    } else {
        let output = service.getEmitOutput(filename, true);
        if (output.emitSkipped) { //checks if output.outputFiles[0]
            return {
                name: filename.replace(dirname, path.join(dirname, "dist", "declarations")),
                content: await fs.readFile(filename, "utf8")
            };
        } else {
            const outputFile = output.outputFiles[0];
            return {
                name: outputFile.name.replace(dirname, path.join(dirname, "dist", "declarations")),
                content: outputFile.text
            };
        }
    }
}
```

evocateur:
Fixes #271 

I had to update some root dependencies (`typescript`, `@babel/preset-typescript`, etc) to make the tests understand `type` exports and whatnot. 580fc7e is the relevant one-liner, in terms of fixing the actual bug.

### DIFF
--- a/.changeset/tiny-bugs-tan.md
+++ b/.changeset/tiny-bugs-tan.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Force emit when generating declaration files
--- a/__fixtures__/typescript-force-dts-emit/.babelrc
+++ b/__fixtures__/typescript-force-dts-emit/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-typescript"]
+}
--- a/__fixtures__/typescript-force-dts-emit/package.json
+++ b/__fixtures__/typescript-force-dts-emit/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "typescript-force-dts-emit",
+  "version": "1.0.0",
+  "main": "dist/typescript-force-dts-emit.cjs.js",
+  "module": "dist/typescript-force-dts-emit.esm.js",
+  "license": "MIT",
+  "private": true,
+  "dependencies": {
+    "@babel/runtime": "^7.8.7",
+    "@reduxjs/toolkit": "^1.3.5"
+  },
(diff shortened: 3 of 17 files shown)


evocateur:
Huh, new type errors (in completely untouched files) from a transitively upgraded version of `rollup`...

I'm barely literate when it comes to TS, I'm not sure if I'm going to be able to make them pass.

mitchellhamilton:
No problem, I'll have a look at the types.

Could you add a changeset?

evocateur:
Whoops, yeah, I totally had that changeset commit sitting on my localhost overnight. :D

evocateur:
Oh, and just to tie the loop explicitly for future archaeologists, the new argument was added in https://github.com/microsoft/TypeScript/commit/850d16e45f3292749c8d9d662b2bd7395ab1d881 (typescript v3.7.2).

evocateur:
Thanks for being so responsive!

Andarist:
I got interested in this change but Im not sure if I understand it completely. Do you happen to know why this has fixed the reported issue? I wouldn't have guessed this would help it by looking at the description of the reported issue. Is it just some TS quirk?

evocateur:
@Andarist I'm very new to TypeScript, generally, but it appears something changed in `typescript@>=3.7.2` that made declaration emission "lazier" than Preconstruct originally expected. I believe it may also be related to `type` import/export, because the presence of this new syntax was directly correlated to the failure I experienced when running `preconstruct build` _without_ this extra argument.

PR #293, opened 2020-07-27
Add experimental ts-morph declaration generator

mitchellhamilton:


ifiokjr:
@mitchellhamilton this look interesting, what are the advantages of `ts-morph`?

mitchellhamilton:
Using it fixes some strange things with the declaration generation in Preconstruct, I haven't been able to figure out how to make Preconstruct's declaration generation completely correct because tsc's API is really confusing.(see the test case in this PR)

Andarist:
So this basically just uses a different API for `.d.ts` generation and doesn't actually have to handle this weird issue in any special way?

packages/cli/src/project.ts
```diff
@@ -24,7 +24,8 @@ export class Project extends Item {
     let config = this._config.___experimentalFlags_WILL_CHANGE_IN_PATCH || {};
     return {
       newEntrypoints: !!config.newEntrypoints,
-      useSourceInsteadOfGeneratingTSDeclarations: !!config.useSourceInsteadOfGeneratingTSDeclarations
+      useSourceInsteadOfGeneratingTSDeclarations: !!config.useSourceInsteadOfGeneratingTSDeclarations,
+      useTSMorphToGenerateTSDeclarations: !!config.useTSMorphToGenerateTSDeclarations,
```
Andarist, line 28: is the plan to remove `useSourceInsteadOfGeneratingTSDeclarations` now then?

mitchellhamilton:
> So this basically just uses a different API for .d.ts generation and doesn't actually have to handle this weird issue in any special way?

Correct. Under the hood, it's still using the TypeScript compiler, ts-morph is just a wrapper that sets it up correctly.

packages/cli/src/project.ts
mitchellhamilton: yep

PR #295, opened 2020-07-27
Run tests on windows

mitchellhamilton:


packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
```diff
@@ -164,7 +165,7 @@ export async function createDeclarationCreator(
               if (
                 !allDeps.has(resolvedModule.resolvedFileName) &&
                 !resolvedModule.isExternalLibraryImport &&
-                resolvedModule.resolvedFileName.includes(dirname)
+                resolvedModule.resolvedFileName.includes(normalizedDirname)
```
Andarist, line 168: @RafalFilipek this should basically fix your issue 😃

RafalFilipek: I'll tattoo this line on my back if it's true :D

Andarist: Better start looking for a tattoo artist 😁

mitchellhamilton:
Thanks for this @Andarist!!

PR #312, opened 2020-08-27
Upgrade dependencies

mitchellhamilton:


.changeset/eight-knives-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
```
evocateur, line 2: This strikes me as a minor release, at a minimum. It's really hard to judge the potential risk of a dependency update when the only thing that ever changes is the patch version.

(Excited for the Rollup v2 upgrade, don't get me wrong!)

PR #322, opened 2020-09-23
Deduplicate babel helpers

### LINKED ISSUE #319: Deduplicate "Babel helpers"
`@rollup/plugin-babel` has a deduplicating logic that works out of the box - preconstruct doesn't though and if one is not using `@babel/transform-runtime`+`@babel/runtime` dependent then preconstruct emits multiple helpers into the bundle. Maybe it would be worth figuring out a similar optimization that `@rollup/plugin-babel` is implementing? For reference see this: https://github.com/preconstruct/preconstruct/issues/297#issuecomment-690964802

mitchellhamilton:
Closes #319


packages/cli/src/worker.ts
```diff
@@ -1,15 +1,68 @@
...
   });
+  if (!config) {
+    return null;
+  }
+  return babel
+    .transformAsync(code, {
+      ...config.options,
+      // note that we're doing this whole thing because we want to add the plugin _before_ user's plugins
+      // so that if they're using @babel/plugin-transform-runtime(which they should be), it'll work
+      plugins: [importHelperPlugin, ...(config.options.plugins || [])],
```
Andarist, line 56: Smart 👍

PR #332, opened 2020-10-19
Cache Babel parsing and code generation

mitchellhamilton:
I'm not 100% sure whether this is a good idea or not but I've been experimenting with caching babel parsing and code generation, I'm not very keen on caching the actual transform bit because of plugins like `babel-plugin-preval` and `babel-plugin-codegen` which often have implicit dependencies on other files/modules which make the caching very unsafe (I mostly think that having a little node script to generate stuff where you might use preval or codegen is better because it doesn't have this caching problem and is kinda just easier to work with, debug and etc.) but even ignoring plugins like those, caching babel transforms has more problems because you can't do things like cache busting based on the version of the plugin or never caching if the plugin isn't in node_modules.


packages/cli/src/worker.ts
```diff
@@ -29,28 +34,76 @@ function importHelperPlugin(babel: typeof import("@babel/core")) {
...
-
-  return babel
-    .transformAsync(code, {
-      caller: {
-        name: "rollup-plugin-babel",
-        supportsStaticESM: true,
-        supportsDynamicImport: true,
+  const { generatorCacheKey,parseCacheKey, options } = resolveOptions({
+    caller: {
+      name: "rollup-plugin-babel",
```
Andarist, line 54: it probably would be a good idea to change this name - it usually doesn't really affect anyone, but might popup in the logs etc so it would be best if one would not think that this is using the original `rollup-plugin-babel` as one might report an issue against wrong repo or smth

packages/cli/src/babel-stuff.ts
```diff
@@ -0,0 +1,119 @@
...
+  parserOpts: babel.ParserOptions;
+  generatorOpts: babel.GeneratorOptions;
+} {
+  const {
+    filename,
+    cwd,
+    filenameRelative = typeof filename === "string"
+      ? path.relative(cwd, filename)
+      : "unknown",
+    sourceType = "module",
```
Andarist, line 19: is this Babel's default? i thought it was `"unambiguous"` but also it didn't matter for me much so I might have been under a wrong impression

mitchellhamilton: The default is "unambiguous" but everything that Preconstruct transforms is a module so I changed this

packages/cli/src/babel-stuff.ts
```diff
@@ -0,0 +1,119 @@
+import * as babel from "@babel/core";
+import path from "path";
+import crypto from "crypto";
+import resolveFrom from "resolve-from";
+
+function normalizeOptions(config: {
```
Andarist, line 6: huh, I assume this is somewhat replicating the logic within Babel itself? is the caching worth the added complexity and having to sync stuff like this with future changes in Babel? this is used both by production builds but also by the internal hook in node and thus during tests etc, right? the latter is a compelling argument for this

mitchellhamilton: Yeah, it is replicating logic within Babel. I'm not super worried about the cost of syncing this with future changes in Babel given that the file this is from hasn't changed in like three years(https://github.com/babel/babel/blob/main/packages/babel-core/src/transformation/normalize-opts.js). I'm mostly concerned about whether this actually yields significant performance improvements and what the disk cost of it is.

> this is used both by production builds but also by the internal hook in node

That's the plan

> thus during tests

This isn't really true at least for Jest because they have their own require implementation which ignores require hooks

Andarist: > This isn't really true at least for Jest because they have their own require implementation which ignores require hooks

Good to know that those 2 ain't fully compatible.

> I'm mostly concerned about whether this actually yields significant performance improvements and what the disk cost of it is.

Right, let me know when u have some numbers - would be interesting to see the results of this experiment.

PR #337, opened 2020-10-28
New dist filenames

### LINKED ISSUE #244: Consider including scope name in the generated filenames
As mentioned in emotion repo - after package renaming our `@emotion/react` will have a confusing file name because it will be shortened to just "react".

I can implement a change if you give a green light on this one.

mitchellhamilton:
Closes #244


.changeset/old-wombats-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Add `newDistFilenames` experimental flag
```
Andarist, line 5: this PR has added `distFilenameStrategy` as well, probably worth mentioning here

mitchellhamilton: `distFilenameStrategy` is behind the `newDistFilenames` flag, will mention it in the changeset that unflags it

Andarist: oh ok - i've missed this when skimming through the implementation, wouldn't it make more sense to have this under experimental key?

mitchellhamilton: The intent is for distFilenameStrategy to be a very well-supported API(once newDistFilenames is unflagged in v2) since people might not want to change their dist filenames for a while/ever

Andarist: Sure, **once** it `newDistFilenames` gets unflagged - then this makes sense to me, but at the moment it's somewhat confusing to have a top-level option that is only enabled with an experimental flag. Not a big deal - but this has caught my completely off guard.

Issue #339, opened 2020-11-05
strange build failure from uncooperating dependency


pkieltyka:
Hello,

please see the following monorepo using preconstruct: https://github.com/pkieltyka/0xsequence.js

to reproduce error:

```
1. yarn install
2. yarn build
```

results in:

https://gist.github.com/pkieltyka/671e55a9024548f38f9442cc01f4c6a6

---

the dependency `ethers` from https://github.com/ethers-io/ethers.js is a monorepo built with typescript and compiled with rollup. The package targets node and browsers, and offers a meta-package "ethers" with most of the sub-packages under "@ethersproject/xxx"

there are many issues with "ethers" to work under preconstruct. Here are the few.

1. First issue is I had to specifically add `@ethersproject/providers` to the dependencies list or preconstruct would complain: https://github.com/pkieltyka/0xsequence.js/blob/master/packages/provider/package.json#L25-L26. This is strange as `@ethersproject/providers` is a dependency of `ethers` already, so I'd hope that preconstruct could infer this list -- if this is a design decision, then please consider this a feature request as it's quite unwieldy in the current form.

2. as you can see in this gist, https://gist.github.com/pkieltyka/671e55a9024548f38f9442cc01f4c6a6 -- the build blows up. It appears as though it's trying to build this with "node" targets, but im targetting a browser. This particular log also has the issue of `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten` warnings, which I'm not sure why they're happening.

---

all of this is to say, that the identical source code builds perfectly fine if I just use `rollup -c` using a very simple rollup config https://gist.github.com/pkieltyka/d11b0fd8f7d83ac581fba39a9bfdbfcb

also FYI, microbundle has almost the identical problems as I tried to use it as well. 

I could just use rollup directly and skip preconstruct, but I like a lot of the extra utilities it offers

pkieltyka:
FYI, https://github.com/ethers-io/ethers.js/issues/998 I came across this issue in the dependency which is being problematic. It seems others are having similar issues with webpack 5, but found ways to get around it. Just wanted to provide that info in case it helps solve around these issues in preconstruct when dependencies become challenging

Andarist:
@mitchellhamilton I'll take a look at this one and report back when I get more info

Andarist:
Do you need a UMD bundle for this? I would suspect that you don't need it - so just to execute `build` successfully you could apply this patch:
```diff
[...]

PR #340, opened 2020-11-06
Allow to build UMD files for packages having dependencies with top-level `this` in ESM files

Andarist:
Relates to https://github.com/preconstruct/preconstruct/issues/339

### DIFF
--- a/packages/cli/src/build/rollup.ts
+++ b/packages/cli/src/build/rollup.ts
@@ -85,9 +85,10 @@ export let getRollupConfig = (
         return;
       }
       switch (warning.code) {
+        case "CIRCULAR_DEPENDENCY":
         case "EMPTY_BUNDLE":
         case "EVAL":
-        case "CIRCULAR_DEPENDENCY":
+        case "THIS_IS_UNDEFINED":
         case "UNUSED_EXTERNAL_IMPORT": {
           break;
         }
--- a/packages/cli/src/build/rollup.ts
+++ b/packages/cli/src/build/rollup.ts
@@ -88,7 +88,6 @@ export let getRollupConfig = (
         case "CIRCULAR_DEPENDENCY":
         case "EMPTY_BUNDLE":
         case "EVAL":
-        case "THIS_IS_UNDEFINED":
         case "UNUSED_EXTERNAL_IMPORT": {
           break;
         }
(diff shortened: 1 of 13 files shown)


packages/cli/src/build/rollup.ts
```diff
@@ -85,9 +85,10 @@ export let getRollupConfig = (
         return;
       }
       switch (warning.code) {
+        case "CIRCULAR_DEPENDENCY":
         case "EMPTY_BUNDLE":
         case "EVAL":
-        case "CIRCULAR_DEPENDENCY":
+        case "THIS_IS_UNDEFINED":
```
Andarist, line 91: as mentioned in that original issue - throwing for this in source files might still be a good idea, this would complicate thing though, currently, this config factory doesn't have access to the list of all monorepo packages so it cant quite distinguish if this has been reported by Rollup for a dependency of a monorepo package

packages/eslint-plugin-format-js-tag/src/index.ts
```diff
@@ -60,7 +81,7 @@ export const rules = {
                 .join("\n"); //+
             //   "\n" +
             //   "".padEnd(node.tag.loc.start.column);
-            if (formatted !== str) {
+            if (!compareLines(str, formatted)) {
```
Andarist, line 84: this means now that the autofixer will reinsert whitespace again and this will potentially result in many meaningless diffs over time, so maybe lack of whitespace-only lines should be enforced rather than just allowed?

packages/cli/src/build/rollup.ts
mitchellhamilton: We can make it so we only ignore it when `type === 'umd'` so the errors will happen for the CJS builds(which will always be built)

packages/eslint-plugin-format-js-tag/src/index.ts
mitchellhamilton: not sure I understand why this has been changed but I'm all for enforcing one kind of formatting

Andarist: I guess this is mainly an issue with my VScode setup that I might even need fixing at some point in time but not sure where this is coming from and I don't have time to tweak this.

Some setting in VSCode (or tools integrated with it) is removing trailing whitespaces for me - this is technically not correct for template literals so that setting has to be some general setting, not knowing about JS-specifics that would turn off this for some lines.

I would expect this kind of a thing to be quite a common setting though so even if I would manage to fix this on my end this might still be a good idea to unify to avoid similar problems for others.

packages/cli/src/build/rollup.ts
Andarist: Hm, I guess this would work. I was previously concerned about the same error not being reported for some build and reported for others which could confuse users, but when I think about this now - it makes sense and shouldn't really be confusing.

WDYT about making this actually only error for cjs if it's built at all times? This would avoid duplicating this error when you build both CJS and ESM

Andarist: > WDYT about making this actually only error for cjs if it's built at all times? This would avoid duplicating this error when you build both CJS and ESM

Ok, I see that this is not quite possible because `"module"` is just built together with `"node-dev"` type

Andarist:
@mitchellhamilton I think this is ready for review and merging now.

mitchellhamilton (approved):

PR #358, opened 2021-01-05
Babel helper things

### LINKED ISSUE #318: Validate that `@babel/plugin-transform-runtime` is configured correctly
It would be nice to add some validation that this plugin has been configured correctly - that `version` parameter was passed in. For reference see this: https://github.com/preconstruct/preconstruct/issues/297#issuecomment-690964802

mitchellhamilton:
Closes #318

cc @Andarist thought you might be interested in this

### DIFF
--- a/packages/cli/src/build/__tests__/basic.ts
+++ b/packages/cli/src/build/__tests__/basic.ts
@@ -549,6 +549,136 @@ test("does not duplicate babel helpers when using @babel/plugin-transform-runtim
   `);
 });
 
+test("does not duplicate babel helpers when not using @babel/plugin-transform-runtime and the helper isn't in the version of @babel/runtime that the user has specified", async () => {
+  const dir = await testdir({
+    "package.json": JSON.stringify({
+      name: "test",
+      main: "dist/test.cjs.js",
+      dependencies: {
+        "@babel/runtime": "*",
+      },
+    }),
+    "babel.config.json": JSON.stringify({
+      presets: [require.resolve("@babel/preset-env")],
+    }),
+    "src/index.js": "import './other'; for (const x of something) {}",
+    "src/other.js": "for (const x of something) {}",
+  });
+  await build(dir);
+  expect(await getDist(dir)).toMatchInlineSnapshot(`
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/test.cjs.dev.js, dist/test.cjs.prod.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    'use strict';
+
+    function _arrayLikeToArray(arr, len) {
+      if (len == null || len > arr.length) len = arr.length;
+
+      for (var i = 0, arr2 = new Array(len); i < len; i++) arr2[i] = arr[i];
+
+      return arr2;
+    }
+
+    function _unsupportedIterableToArray(o, minLen) {
+      if (!o) return;
+      if (typeof o === "string") return _arrayLikeToArray(o, minLen);
+      var n = Object.prototype.toString.call(o).slice(8, -1);
+      if (n === "Object" && o.constructor) n = o.constructor.name;
+      if (n === "Map" || n === "Set") return Array.from(n);
+      if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)) return _arrayLikeToArray(o, minLen);
+    }
+
+    function _createForOfIteratorHelper(o) {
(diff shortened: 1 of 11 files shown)


packages/cli/src/build/config.ts
```diff
@@ -57,6 +57,11 @@ function getGlobal(project: Project, name: string) {
   }
 }
 
+const babelHelperId = /@babel\/runtime(|-corejs[23])\/helpers\//;
+
+const interop = (id: string | null): "auto" | "default" =>
```
Andarist, line 62: TIL that this can be a function

packages/cli/src/build/__tests__/basic.ts
```diff
@@ -366,6 +549,66 @@ test("does not duplicate babel helpers when using @babel/plugin-transform-runtim
...
+    }),
+    "src/index.js": "import './other'; for (const x of something) {}",
+    "src/other.js": "for (const x of something) {}",
+  });
+  await build(dir);
+  expect(await getDist(dir)).toMatchInlineSnapshot(`
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/test.cjs.dev.js, dist/test.cjs.prod.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    'use strict';
+
+    var _createForOfIteratorHelper = require('@babel/runtime/helpers/createForOfIteratorHelper');
```
Andarist, line 572: shouldnt u add a very similar test that would check that this helper has been inlined rather than required?

packages/cli/src/build/config.ts
Andarist: Have you checked how this will behave in Babel 8? I know that more than this would have to be re-checked to check against Babel 8 compatibility but maybe worth to be future-proof here. 

In Babel 8 there won't be separate cjs/esm versions of helpers as they have started to use conditional exports. You would have to check if `interop: "default"` works with Babel 8 though - there might have been changes around this (but they might have also left it as is - I don't recall, I think Nicolo has wanted to support `.default`-less requires in node so probably this has not been touched but... well, worth re-checking).

packages/cli/src/rollup-plugins/flow-and-prod-dev-entry.ts
```diff
@@ -49,6 +49,7 @@ export default function flowAndNodeDevProdEntry(
       }
 
       if (
+        source.startsWith("\0") ||
```
Andarist, line 52: wait - is `\0babelWhatever` added to the top of a generated virtual source code? 🤔 how does it work? kinda blurry for me in a rush

Andarist:
How do u ensure that this doesn't clash with configurations using the transform plugin explicitly? 🤔

packages/cli/src/rollup-plugins/flow-and-prod-dev-entry.ts
mitchellhamilton: `source` is referring to the unresolved module specifier, not source code.

packages/cli/src/build/__tests__/basic.ts
mitchellhamilton: As in when the version of @babel/runtime doesn't include that helper? That test already exists before this PR - https://github.com/preconstruct/preconstruct/blob/b9e91373985b967daa6440ba15970e3e8dd776d2/packages/cli/src/build/__tests__/basic.ts#L238-L367

mitchellhamilton:
> How do u ensure that this doesn't clash with configurations using the transform plugin explicitly? 🤔

https://github.com/preconstruct/preconstruct/blob/2f63b87185cd0f0687b26700634bd1d4f4174846/packages/cli/src/worker.ts#L9-L14

(that plugin is always last)

packages/cli/src/rollup-plugins/flow-and-prod-dev-entry.ts
Andarist: Oh, gotcha.

packages/cli/src/build/__tests__/basic.ts
Andarist: a similar test - yes, but without `@babel/plugin-transform-runtime` being used

packages/cli/src/build/config.ts
mitchellhamilton: Given [the PR about it](https://github.com/babel/babel/pull/12295) isn't merged yet, I'm not overly concerned about it. Preconstruct has a direct dependency on @babel/core@7 rn anyway.

Andarist (approved):
From whatever my ✅ is worth here:

✅ ✅ ✅ ✅ ...

Great attention to little things as always 👍

PR #366, opened 2021-02-02
Fixed destination paths of generated `.d.ts` files in TypeScript projects on Windows

Andarist:
cc @RafalFilipek

### DIFF
--- a/packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
+++ b/packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
@@ -183,8 +183,8 @@ export async function createDeclarationCreator(
       if (filename.endsWith(".d.ts")) {
         return {
           name: filename.replace(
-            dirname,
-            path.join(dirname, "dist", "declarations")
+            normalizePath(dirname),
+            normalizePath(path.join(dirname, "dist", "declarations"))
           ),
           content: await fs.readFile(filename, "utf8"),
         };
@@ -192,8 +192,8 @@ export async function createDeclarationCreator(
       let output = service.getEmitOutput(filename, true, true);
       return {
         name: output.outputFiles[0].name.replace(
-          dirname,
-          path.join(dirname, "dist", "declarations")
+          normalizePath(dirname),
+          normalizePath(path.join(dirname, "dist", "declarations"))
         ),
         content: output.outputFiles[0].text,
       };
--- a/.changeset/eight-parrots-film.md
+++ b/.changeset/eight-parrots-film.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
```diff
@@ -114,7 +114,7 @@ export async function createDeclarationCreator(
   // if the tsconfig _isn't_ in the package directory though, it's probably fine to memoize it
   // since it should just be a root level tsconfig
   let { service, options, program } = await (normalizePath(configFileName) ===
-  normalizePath(path.join(dirname, "tsconfig.json"))
+  path.join(dirname, "tsconfig.json")
```
Andarist, line 117: `path.join` already calls `path.normalize` under the hood so this was not quite needed

packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
```diff
@@ -192,7 +192,7 @@ export async function createDeclarationCreator(
       let output = service.getEmitOutput(filename, true, true);
       return {
         name: output.outputFiles[0].name.replace(
-          dirname,
+          normalizePath(dirname),
```
Andarist, line 195: this couldn't be replaced correctly because the `.name` was already normalized the replacing pattern was also normalized, but the input string (`dirname`) was not

packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
mitchellhamilton: path.normalize !== normalize-path. [normalize-path](https://www.npmjs.com/package/normalize-path) unixifies paths, path.normalize transforms paths into the platform path separator

Andarist: Well, sure - `path.normalize` for example preserves a trailing slash **but** this quite can't if we are dealing with a `path.join` results. I honestly can't name what difference does it make here - maybe there is something obscure that I'm forgetting but I've done some tests and looked into `path.normalize` & `path.join` implementations and I think this is safe to be done

mitchellhamilton: normalize-path always returns paths with `/`, path.normalize will return paths with `\` on Windows so with the change on this line, the two will never be equal on Windows

Andarist: Ah, right - I was looking into the code that was executing on my machine but I've missed that the `path` module switches the whole implementations: 
https://github.com/nodejs/node/blob/fe43bd8ddd16a614ba157c57c438cd2f206c5603/lib/path.js#L1420

mitchellhamilton (approved):
Thanks!!

Issue #360, opened 2021-01-18
Allow Babel config based on babelrc files


Andarist:
Currently, Babel only knows about the root of the project:
https://github.com/preconstruct/preconstruct/blob/764bb9b05851bf88ecc62179628a8d1ff1ac26f8/packages/cli/src/worker.ts#L48-L49
and thus it can't resolve non-root configs. It would be great to have `babelrcRoots` configured properly for packages or at least add some docs that this is a limitation of preconstruct (or a deliberate design decision). If this won't get supported a helpful validation would also be great.

mitchellhamilton:
Hmmm, that's interesting, this wasn't a deliberate decision. I'll have a look

mitchellhamilton:
As discussed in https://github.com/preconstruct/preconstruct/pull/371#issuecomment-785438082, this is now intentionally not done, you can use `babelrcRoots` in the root Babel config

PR #371, opened 2021-02-10
Allow Babel config based on babelrc files

RafalFilipek:
Hi!

There is only one change in `worker.ts` but I need help with tests. I made some and it works. I'm not sure how to test `dev` mode.

I'll appreciate your feedback :)

Fix for #360.


RafalFilipek:
@mitchellhamilton I can see from your timeline that you're quite busy, but I wondered if there is anything I can add to this MR?

mitchellhamilton:
I've been thinking about this and I think I'd like to not add `babelrcRoots: true` for a couple reasons and instead recommend that people add `babelrcRoots: true` to their Babel config if they need it since

- Most people don't need it
- It adds more fs calls
- It would make preconstruct dev usage more complicated if it was configured in Preconstruct itself since the require hook and `@preconstruct/next` would need it but more importantly, it would mean another thing people would have to set in other places where they currently just have to set `cwd` or `rootMode: 'upward-optional'`

Andarist:
TIL that `babelrcRoots` can be configured in the `babel.config.js`

RafalFilipek:
True :)

Right now I have `babel.config.js` with `babelrcRoots` and it works super cool with packages and apps. I have 2 NextJS app and 4 packages, each with separated `.babelrc.js`. All possible combinations work just fine.

Issue #375, opened 2021-02-25
Preconstruct watch not working properly with entrypoints


dbismut:
Hi, I'm currently building this monorepo library https://github.com/dbismut/preconstruct-repro and chose `preconstruct` as the build tool. Fantastic experience so far I love it.

I'm having a bug though, with `preconstruct watch` modifying my own code. I've made a minimum repro showing the issue:
https://github.com/dbismut/preconstruct-repro

Just run `yarn preconstruct watch` and look at what happens inside the file inside `src/utils/index.js`.

```js
// what I wrote
export { fn2 } from "../data";

// how preconstruct watch alters it:

// at first:
export { f as fn2 } from '../../dist/data-11edc36a.esm.js';
// and then:
function fn2() {
  return 3;
}

export { fn2 };
```
 
I'm guessing this has to do with the secondary entry point exporting a file which is also consumed by the main entry point in `src/index.js`?

EDIT: forgot to mention that `preconstruct build && preconstruct watch` solves it in my case.

Andarist:
Thanks for the report - I might take a look at it soon-ish but no hard promises

dbismut:
Thanks @Andarist, just spotted your answer now. Btw, in dev mode, I have an issue with relative imports from the esm.js source file that can't be resolved. Is this something that you're aware of? I can also file an issue for this.

Andarist:
Please create a repro case for that other case and open a new issue for it. As to this issue here - I've managed to prepare a fix: https://github.com/preconstruct/preconstruct/pull/376

Issue #306, opened 2020-08-19
Error: EINVAL: invalid argument, unlink '<FILENAME>'


sarahbethfederman:
I am getting a non-deterministic error when trying to run `preconstruct build`. It errors on a different file name each time. When I try to run the build command again directly afterwards it either repeats the error with a different file 1-4 times or builds successfully. Here's an example: 
```
sarahfederman@syd-sfederman-l lui-2 % yarn preconstruct build
yarn run v1.22.4
$ /Users/sarahfederman/Documents/lui-2/node_modules/.bin/preconstruct build
🎁 info building bundles!
🎁 error { [Error: EINVAL: invalid argument, unlink '/Users/sarahfederman/Documents/lui-2/utils/doc-utils/dist/doc-utils.esm.js']
🎁 error   errno: -22,
🎁 error   code: 'EINVAL',
🎁 error   syscall: 'unlink',
🎁 error   path:
🎁 error    '/Users/sarahfederman/Documents/lui-2/utils/doc-utils/dist/doc-utils.esm.js' }
🎁 info If want to learn more about the above error, check https://preconstruct.tools/errors
🎁 info If the error is not there and you want to learn more about it, open an issue at https://github.com/preconstruct/preconstruct/issues/new
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Seems like there's maybe a check missing somewhere?

mitchellhamilton:
I think I've found the issue (https://github.com/preconstruct/preconstruct/commit/d56018d6793d7298912bf2e4dc9fd92c981e7e9f), could you try `@preconstruct/cli@1.1.23`?

sarahbethfederman:
Seems to have worked!

PR #376, opened 2021-03-01
Fixed watch not cleaning up all entrypoints correctly

Andarist:
fixes https://github.com/preconstruct/preconstruct/issues/375

### DIFF
--- a/packages/cli/src/build/utils.ts
+++ b/packages/cli/src/build/utils.ts
@@ -1,7 +1,40 @@
+import { Project } from "../project";
+import { isTsPath } from "../rollup-plugins/typescript-declarations";
+import { writeDevTSFile } from "../dev";
+import * as fs from "fs-extra";
+import path from "path";
+
 export function getDevPath(cjsPath: string) {
   return cjsPath.replace(/\.js$/, ".dev.js");
 }
 
 export function getProdPath(cjsPath: string) {
   return cjsPath.replace(/\.js$/, ".prod.js");
+}
+
+export async function cleanProjectBeforeBuild(project: Project) {
+  await Promise.all(
+    project.packages.map(async (pkg) => {
+      await Promise.all([
+        fs.remove(path.join(pkg.directory, "dist")),
+        ...pkg.entrypoints
+          .filter((entrypoint) => entrypoint.name !== pkg.name)
+          .map((entrypoint) => {
+            return fs.remove(path.join(entrypoint.directory, "dist"));
+          }),
+      ]);
+
+      await Promise.all(
+        pkg.entrypoints.map(async (entrypoint) => {
+          if (isTsPath(entrypoint.source)) {
+            await fs.mkdir(path.join(entrypoint.directory, "dist"));
+            await writeDevTSFile(
+              entrypoint,
+              await fs.readFile(entrypoint.source, "utf8")
+            );
+          }
+        })
+      );
+    })
+  );
 }
--- a/packages/cli/src/build/watch.ts
+++ b/packages/cli/src/build/watch.ts
@@ -4,13 +4,13 @@ import { watch } from "rollup";
 import chalk from "chalk";
 import path from "path";
 import ms from "ms";
-import * as fs from "fs-extra";
 import { getRollupConfigs } from "./config";
 import { Aliases, getAliases } from "./aliases";
 import { success, info } from "../logger";
 import { successes } from "../messages";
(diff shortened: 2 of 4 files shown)


packages/cli/src/build/watch.ts
```diff
@@ -18,7 +18,13 @@ function relativePath(id: string) {
 
 async function watchPackage(pkg: Package, aliases: Aliases) {
   const _configs = getRollupConfigs(pkg, aliases);
-  await fs.remove(path.join(pkg.directory, "dist"));
+
+  await Promise.all(
+    pkg.entrypoints.map((entrypoint) =>
```
Andarist, line 23: based on https://github.com/preconstruct/preconstruct/blob/aa54f62e55ce47e6e2c7af59a1ff1f33fbda127e/packages/cli/src/build/rollup.ts#L61-L65

mitchellhamilton (approved):
Thanks!

packages/cli/src/build/utils.ts
```diff
@@ -1,7 +1,40 @@
...
   return cjsPath.replace(/\.js$/, ".prod.js");
 }
+
+export async function cleanProjectBeforeBuild(project: Project) {
+  await Promise.all(
+    project.packages.map(async (pkg) => {
+      await Promise.all([
+        fs.remove(path.join(pkg.directory, "dist")),
+        ...pkg.entrypoints
+          .filter((entrypoint) => entrypoint.name !== pkg.name)
```
Andarist, line 21: is this filter needed? the piece that I've based my code on didn't do any of this - it just has used `pkg.entrypoints` as the base for computing the paths:
https://github.com/preconstruct/preconstruct/blob/aa54f62e55ce47e6e2c7af59a1ff1f33fbda127e/packages/cli/src/build/rollup.ts#L57-L67

packages/cli/src/build/utils.ts
```diff
@@ -1,7 +1,40 @@
...
+        pkg.entrypoints.map(async (entrypoint) => {
+          if (isTsPath(entrypoint.source)) {
+            await fs.mkdir(path.join(entrypoint.directory, "dist"));
+            await writeDevTSFile(
+              entrypoint,
+              await fs.readFile(entrypoint.source, "utf8")
+            );
+          }
+        })
+      );
```
Andarist, line 37: is this required for the watch script? can it result in some unwanted side-effects? why this is needed for the build script?

packages/cli/src/build/utils.ts
mitchellhamilton: That code is doing a different thing. Here we want to clear all the entrypoint dists _and_ the root dist directory because the root dist directory will contain shared chunks and the TS declarations, and the filter is necessary because the root is _sometimes_ an entrypoint so we don't try to delete the same thing at the same time (#306). I suppose it's not technically necessary to remove the root one for watch since there shouldn't be symlinks there if it's not an entrypoint but it's nice for consistency with build and maybe if the root _was_ an entrypoint, preconstruct dev was run, it was removed as an entrypoint and preconstruct watch was run

packages/cli/src/build/utils.ts
mitchellhamilton: It's needed because we build everything in parallel so when doing a build, the TypeScript compiler needs to be able to resolve packages before they're built.

Andarist: I guess this doesn't create problems that much so it's OK. Out of curiosity - have you considered using `paths` options in the tsconfig (the one created/loaded programmatically and passed to the TS API)? Or have you not wanted to increase the complexity there?

packages/cli/src/build/utils.ts
Andarist: So the summary would be that the root isn't always an entrypoint but the root dist is always created regardless. Good to know - thanks for the explanation 👍

packages/cli/src/build/utils.ts
mitchellhamilton: I tried but couldn't get it to work with paths ¯\\\_(ツ)_/¯

packages/cli/src/build/utils.ts
mitchellhamilton: > the root dist is always created regardless

it will be created in _some_ cases(currently if someone is using TypeScript or has multiple entrypoints with common modules between them), it won't always be created.

PR #383, opened 2021-03-22
Add keepDynamicImportAsDynamicImportInCommonJS experimental flag

mitchellhamilton:
This is part 1 of Node ESM support in Preconstruct, solving the problem of "I want to ship a package that works from imports or requires but I want to depend on an ESM only package and I'm okay with only being able to get the dependency asynchronously"

cc @Andarist

### DIFF
--- a/.changeset/hip-steaks-swim.md
+++ b/.changeset/hip-steaks-swim.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Add `keepDynamicImportAsDynamicImportInCommonJS` experimental flag to allow importing Node ESM packages in Preconstruct packages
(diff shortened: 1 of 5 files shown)


.changeset/hip-steaks-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Add `keepDynamicImportAsDynamicImportInCommonJS` experimental flag to allow importing Node ESM packages in Preconstruct packages
```
Andarist, line 5:
```suggestion
Add `keepDynamicImportAsDynamicImportInCommonJS` experimental flag to allow importing Node ESM-only packages in Preconstruct packages. Useful when one wants to ship a package that can be both imported and required and when having a dependency that doesn't have a CommonJS distribution. 

Note that `import()` is asynchronous so it's won't be possible to get access to such ESM-only dependency synchronously.
```

Andarist (approved):

PR #386, opened 2021-03-23
Add details to monorepo tutorial Babel install

with-heart:
This PR adds a few extra details to `tutorials/monorepo.mdx` to increase clarity/reduce confusion:

- added `@babel/core` to the list of `yarn add` packages, since it's mentioned above but not included in the actual command
- added `-W` flag to `yarn add` command, to make it clear that these packages are being installed to the root `package.json`
- added a link to the "Building TypeScript packages" guide, with a note that `@babel/preset-typescript` should be _added_ to the `presets` list


mitchellhamilton (approved):
Thanks!!

PR #392, opened 2021-04-24
Document TypeScript compiler configuration

jrolfs:
Document the `isolatedModules` TypeScript compiler/project option in the TypeScript guide. I also added an entry to the errors page, let me know if you think that makes sense.

I did come across some `An unhandled Rollup error occurred: *` when initially troubleshooting this so I could look into adding a friendlier error message if you think that'd be worthwhile.

### DIFF
--- a/site/src/pages/errors.mdx
+++ b/site/src/pages/errors.mdx
@@ -24,4 +24,4 @@ Preconstruct requires that all entrypoints in a package have the same fields so 
 
 ## '{export-a}' is not exported by {module-a}, imported by {module-b}
 
-If you are building a TypeScript package, chances are you're re-exporting a type in a way that's not compatible with `@babel/preset-typescript`. See [Building TypeScript packages](/guides/building-typescript-packages) for more information.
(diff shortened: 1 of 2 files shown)


site/src/pages/errors.mdx
```diff
@@ -21,3 +21,7 @@ Preconstruct requires that all imports in a package must either packages that ar
 ## {entrypoint-name-a} has a {field-name} build but {entrypoint-name-b} does not have a {field-name} build. Entrypoints in a package must either all have a particular build type or all not have a particular build type.
 
 Preconstruct requires that all entrypoints in a package have the same fields so that it can guarantee that common modules between entrypoints will never be instantiated twice within an environment.
+
+## '{export-a}' is not exported by {module-a}, imported by {module-b}
+
+If you are building a TypeScript package, chances are you're re-exporting a type in a way that's not compatible with `@babel/preset-typescript`. See [Building TypeScript packages](/guides/building-typescript-packages) for more information.
```
mitchellhamilton, line 27:
```suggestion
If you are building a TypeScript package, chances are you're re-exporting a type in a way that's not compatible with `@babel/preset-typescript`. See [Building TypeScript packages](/guides/building-typescript-packages) for more information. To fix it, you should use [type-only imports and exports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)
```

mitchellhamilton (approved):
Thanks!!

> I did come across some An unhandled Rollup error occurred: * when initially troubleshooting this so I could look into adding a friendlier error message if you think that'd be worthwhile.

That would be great!

PR #393, opened 2021-05-04
feat: generate d.ts.map files

with-heart:
This PR makes a change to handle `d.ts.map` files generated when the `tsconfig.declarationMap` option is enabled.

When this option is enabled, `LanguageService.getEmitOutput` outputs an array of the `d.ts.map` and `d.ts` files and their output, so I changed `getDeclarationFile` to `getDeclarationFiles`, and modified `generateBundle` to map over the files returned from it.

I guess this could technically be considered a bug since only the `d.ts.map` files are outputted currently, but feature sounds nicer :P

### DIFF
--- a/packages/cli/src/build/__tests__/basic.ts
+++ b/packages/cli/src/build/__tests__/basic.ts
@@ -101,6 +101,86 @@ test("typescript thing", async () => {
   await snapshotDirectory(path.join(tmpPath, "dist"), { files: "all" });
 });
 
+test("typescript declarationMap", async () => {
+  let dir = await testdir({
+    "package.json": JSON.stringify({
+      name: "typescript-declarationMap",
+      main: "dist/typescript-declarationMap.cjs.js",
+      module: "dist/typescript-declarationMap.esm.js",
+
+      dependencies: {
+        "@babel/runtime": "^7.8.7",
+      },
+
+      devDependencies: {
+        typescript: "^3.8.3",
+      },
+    }),
+    ".babelrc": JSON.stringify({
+      presets: [require.resolve("@babel/preset-typescript")],
+    }),
+    node_modules: {
+      kind: "symlink",
+      path: repoNodeModules,
+    },
+    "tsconfig.json": JSON.stringify(
+      {
+        compilerOptions: {
+          target: "esnext",
+          module: "esnext",
+          declarationMap: true,
+          jsx: "react",
+          isolatedModules: true,
+          strict: true,
+          moduleResolution: "node",
+          esModuleInterop: true,
+          noEmit: true,
+        },
+      },
+      null,
+      2
+    ),
+    "src/index.ts": ts`
+                      export const thing = "wow" as const;
+                    `,
+  });
+
+  await build(dir);
+  await expect(getDist(dir)).resolves.toMatchInlineSnapshot(`
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/declarations/src/index.d.ts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          export declare const thing: "wow";
+          //# sourceMappingURL=index.d.ts.map
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/declarations/src/index.d.ts.map ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          {"version":3,"file":"index.d.ts","sourceRoot":"","sources":["index.ts"],"names":[],"mappings":"AAAA,eAAO,MAAM,KAAK,OAAiB,CAAC"}
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/typescript-declarationMap.cjs.d.ts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          export * from "./declarations/src/index";
+
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/typescript-declarationMap.cjs.dev.js, dist/typescript-declarationMap.cjs.prod.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          'use strict';
+
+          Object.defineProperty(exports, '__esModule', { value: true });
+
+          const thing = "wow";
+
+          exports.thing = thing;
+
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/typescript-declarationMap.cjs.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          'use strict';
+
+          if (process.env.NODE_ENV === "production") {
+            module.exports = require("./typescript-declarationMap.cjs.prod.js");
+          } else {
+            module.exports = require("./typescript-declarationMap.cjs.dev.js");
+          }
+
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/typescript-declarationMap.esm.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          const thing = "wow";
+
+          export { thing };
+
+        `);
+});
+
 test("process.env.NODE_ENV reassignment", async () => {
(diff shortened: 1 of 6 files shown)


packages/cli/test-utils/index.ts
```diff
@@ -395,7 +395,7 @@ async function readNormalizedFile(filePath: string): Promise<string> {
   let content = await fs.readFile(filePath, "utf8");
   // to normalise windows line endings
   content = content.replace(/\r\n/g, "\n");
-  if (/\.map$/.test(filePath)) {
+  if (/(?<!(\.d\.ts))\.map$/.test(filePath)) {
```
with-heart, line 398: Added a negative lookbehind so this won't match `d.ts.map` files.

packages/cli/src/rollup-plugins/typescript-declarations/index.ts
```diff
@@ -22,12 +22,14 @@ export default function typescriptDeclarations(pkg: Package): Plugin {
...
         [...deps].map(async (dep) => {
-          let { name, content } = await creator.getDeclarationFile(dep);
-          srcFilenameToDtsFilenameMap.set(normalizePath(dep), name);
-          this.emitFile({
-            type: "asset",
-            fileName: path.relative(opts.dir!, name),
-            source: content,
+          let files = await creator.getDeclarationFiles(dep);
+          files.forEach(({ name, content }) => {
+            srcFilenameToDtsFilenameMap.set(normalizePath(dep), name);
```
Andarist, line 27: this has the potential of overriding the `.d.ts` value with `.d.ts.map`. I have no idea how this map is used so not sure how this can affect the program but looks like maybe something that should be addressed

mitchellhamilton: This should only set it for the `.d.ts` one, it used used futher down in this file to determine what the declaration re-export should point to.

mitchellhamilton: (i suspect it happens to work right now because the dts map comes first in the array but this should still be fixed)

mitchellhamilton:
Have you found that with this change, the dts maps actually work? I just tried building a TS project with Preconstruct from this branch and it generated the dts maps but go to definition pointed to the declaration files rather than the source, my understanding is the point of declaration file source maps is that go to definition will point to source files rather than the declaration files, is that right? (note you can use a local version of Preconstruct in another project by doing `node ../path/to/preconstruct/repo/package/cli/bin.js build`) Also, would I be correct in understanding that using declaration maps requires that you ship your source files to npm because the source files are not included in the source maps?

mitchellhamilton:
Just tried a thing locally and it seems like if sourceRoot is set to the directory that the source file was in, things work as expected. There's probably some TS config stuff that we could do to make TS output the right thing but I'm not sure if that will be possible with the one TS config/language service for multiple packages thing that Preconstruct does. I think an easier and probably more reliable solution would be to overwrite the sourceRoot of the generated map in Preconstruct.

packages/cli/src/build/__tests__/basic.ts
```diff
@@ -101,6 +101,86 @@ test("typescript thing", async () => {
...
+                    `,
+  });
+
+  await build(dir);
+  await expect(getDist(dir)).resolves.toMatchInlineSnapshot(`
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/declarations/src/index.d.ts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          export declare const thing: "wow";
+          //# sourceMappingURL=index.d.ts.map
+          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/declarations/src/index.d.ts.map ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+          {"version":3,"file":"index.d.ts","sourceRoot":"","sources":["index.ts"],"names":[],"mappings":"AAAA,eAAO,MAAM,KAAK,OAAiB,CAAC"}
```
mitchellhamilton, line 154: For this one as an example, the sourceRoot should look like this I believe:
```suggestion
          {"version":3,"file":"index.d.ts","sourceRoot":"../../../src","sources":["index.ts"],"names":[],"mappings":"AAAA,eAAO,MAAM,KAAK,OAAiB,CAAC"}
```

packages/cli/src/rollup-plugins/typescript-declarations/index.ts
Andarist: > (i suspect it happens to work right now because the dts map comes first in the array but this should still be fixed)

Yep, that was my suspicion as well.

Andarist:
> Also, would I be correct in understanding that using declaration maps requires that you ship your source files to npm because the source files are not included in the source maps?

I believe so - and was just thinking (before reading your comment) that an issue about adding a validation for this should be created (imho not worth overcomplicating this PR).

> There's probably some TS config stuff that we could do to make TS output the right thing but I'm not sure if that will be possible with the one TS config/language service for multiple packages thing that Preconstruct does

How does it work then that relative paths are correct for the generated `.d.ts`? :thiking_face:

packages/cli/src/rollup-plugins/typescript-declarations/create-generator.ts
```diff
@@ -178,26 +178,28 @@ export async function createDeclarationCreator(
       searchDeps(new Set(resolvedEntrypointPaths));
       return allDeps;
     },
-    getDeclarationFile: async (
+    getDeclarationFiles: async (
       filename: string
-    ): Promise<{ name: string; content: string }> => {
+    ): Promise<{ name: string; content: string }[]> => {
```
Andarist, line 183: since there are concerns about how this gets consumed later, maybe we should refactor this to return:
```ts
interface DeclarationFile {
  name: string;
  content: string;
}

Promise<{ types: DeclarationFile, map?: DeclarationFile }>
```

mitchellhamilton:
> How does it work then that relative paths are correct for the generated .d.ts? 🤔

Because the only relative paths are to other source files in the package which will have .d.ts files generated into the dist.

mitchellhamilton:
> I believe so - and was just thinking (before reading your comment) that an issue about adding a validation for this should be created (imho not worth overcomplicating this PR).

Agreed

Andarist:
> Because the only relative paths are to other source files in the package which will have .d.ts files generated into the dist.

Ah, OK - thank you for the clarification. `.d.ts` never link back to anything relative outside of `src` and the structure of the generated `dist/declarations` is the same as in the `src` so it just works.

with-heart:
> Also, would I be correct in understanding that using declaration maps requires that you ship your source files to npm because the source files are not included in the source maps?

That's a good question and it sounds correct, but I'm honestly not sure. I'm working on [converting Chakra UI to `preconstruct`](https://github.com/chakra-ui/chakra-ui/pull/3956) and we had `declarationMap` enabled but I never noticed until working on the conversion. 

Testing it in an existing project using Chakra (verified that `d.ts.map` files exist), it appears that we would need to include `src` in order for it map to that file. Right now, it's still showing the `d.ts` definition for Go To Declaration.

Interesting enough, when compiling them using `babel`, `sourceRoot` is `""` and `sources` is `["../../src/file.tsx"]`. Seems weird that `sourceRoot` is never used. Not sure what the implications would be.

with-heart:
I believe I've resolved the `sourceRoot` and `srcFilenameToDtsFilenameMap` issues. We're now processing `types` and `map` separately, with the `sourceRoot` value of the `d.ts.map` file being updated to the correct path. I tested it locally with `yarn link` + Chakra and Go To Declaration now goes directly to the `src` file rather than `d.ts`.

There's a failure on the Windows build for this change that I'm not sure how to resolve, since paths are calculated differently. Suggestions?

Andarist:
> There's a failure on the Windows build for this change that I'm not sure how to resolve, since paths are calculated differently. Suggestions?

You might want to expand on the normalizing logic here: https://github.com/preconstruct/preconstruct/pull/393/files#diff-2541fa35d8c597774265114a18c6005320fbe05f8c14c93462f80b8afb2cb904R398

mitchellhamilton:
You should normalize the path to unix style (with the `normalize-path` package)

Andarist:
Just wondering - maybe we don't need to normalize the snapshot output and we could just normalize the paths in the map that is written to disk? It would only make sense if an IDE on Windows would be capable of processing unix paths - since maps can be generated on a different machine, right?

mitchellhamilton:
Yes, sorry I should have been more clear, we should normalize the path we write to the .d.ts map, not normalize it in the snapshot.

with-heart:
I think we should be good to go!

Andarist (approved):

mitchellhamilton (approved):
Thanks!

PR #409, opened 2021-08-30
feat: allow only esm builds

prisis:
- [ ] change typescript types file generating

### DIFF
--- a/__fixtures__/no-main/package.json
+++ b/__fixtures__/no-main/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "no-main",
+  "version": "1.0.0",
+  "module": "dist/no-main.esm.js",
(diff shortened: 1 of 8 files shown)


prisis:
@mitchellhamilton hey :), are you fine with this proposal?

If yes, than i would finish it.

mitchellhamilton:
I'm assuming by ESM only, you mean that you'd have an `exports` field as well so it works in Node? Supporting ESM only builds is definitely something that I'd like Preconstruct to support but it needs quite a bit of design and I'd like to wait for TypeScript's Node ESM support first

prisis:
Yeah, exports or the module key 

Okay than I will do the changes for me

Thanks for the library :)

PR #413, opened 2021-09-09
fix logger receiveing `object` type `message`

namjul:
I encountered the error `TypeError: message.split is not a function` when using jsx without having `@babel/preset-react` defined.

It looks to me that `message` can be an `object` and casting it to string solved the issue for me.


mitchellhamilton (approved):
Thanks!

PR #422, opened 2021-11-17
upgrade repo  to eslint8 ecosystem

gwyneplaine:
This is currently blocking upgrade to eslint 8 for keystone https://github.com/keystonejs/keystone/pull/6769


packages/eslint-plugin-format-js-tag/src/index.ts
```diff
@@ -21,8 +21,7 @@ export const rules = {
     defaultOptions: [],
     meta: {
       docs: {
-        category: "Stylistic Issues",
-        description: "",
+        description: "Stylistic Issues",
```
gwyneplaine, line 24: @mitchellhamilton  category property removed from docs config as of this PR https://github.com/typescript-eslint/typescript-eslint/pull/3800, replaced the description in lieu of not knowing where else to put it

mitchellhamilton:
```suggestion
        description: "Formats `js`, `ts` and `tsx` template tags with Prettier",
```

mitchellhamilton:
It looks like TS is failing because the typescript-eslint packages are depending on types that are new in TS 4.5, given 4.5 should be released ~tomorrow, I'll update TS in this repo when that happens and then merge this

mitchellhamilton (approved):

PR #441, opened 2022-01-14
Limit `fast-glob` version range because of issue

MurzNN:
`fast-glob` versions after 3.2.7 has an issue https://github.com/mrmlnc/fast-glob/issues/351
So it's good to temporary limit the version as workaround, until that issue is fixed.


MurzNN:
Closing in favor of https://github.com/preconstruct/preconstruct/pull/442, because `fast-glob` fixed this issue in last release.

PR #481, opened 2022-07-18
Fix unresolveable packages

### LINKED ISSUE #480: Package is not specified in dependencies error for UMD builds
Hey, I'm using Preconstruct for a monorepo of a client and I'm trying to add UMD builds, but for packages where I use direct imports I'm getting the "package is not specified in dependencies or peerDependencies" error. For example:

```
"@atomico/hooks/use-slot" is imported by "src/hooks/useChildren.ts" but the package is not specified in dependencies or peerDependencies
```

The `@atomico/hooks` package is using `exports` in package.json and I can only use direct imports like that.

CJS and ESM builds work fine.

efoken:
This PR fixes #480 

I updated the @rollup/plugin-node-resolve plugin to make direct imports resolveable.

### DIFF
--- a/.changeset/tough-ligers-kick.md
+++ b/.changeset/tough-ligers-kick.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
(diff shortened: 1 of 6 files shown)


mitchellhamilton (approved):
Thanks!

PR #485, opened 2022-07-19
feat: support `"type": "module"`

### LINKED ISSUE #471: Preconstruct ignores "type": "module"
Regardless of whether or not one of my package's package.json contains `"type": "module"`, Preconstruct always makes `"main"` point to the CommonJS bundle. However, bundlers generally expect a dependency with `"type": "module"` to be an ESM package whose package.json `"main"` points to the ESM build. An ESM package shouldn't have a CommonJS bundle.

As an example, [this is how Parcel works](https://parceljs.org/features/targets/#library-targets).

> Library targets automatically output either native ES modules or CommonJS depending on the target.
> 
> `main` – by default, outputs CommonJS. If the `.mjs` extension is used, or the `"type": "module"` field is specified, then an ES module is output instead.
> `module` – outputs an ES module.
> `browser` – a browser-specific override of the `main` field. Outputs CommonJS.

I'm running into this issue because one of my packages uses `react-markdown` which is an ESM dependency, but in my unwanted CommonJS bundle, it's being `require`d.

The following package.json snippet should be invalid:

```
"type": "module",
"main": "dist/my-package.cjs.js",
```
and should be fixed as
```
"type": "module",
"main": "dist/my-package.esm.js",
```

And
```
"type": "module",
"main": "dist/my-package.cjs.js",
"module": "dist/my-package.cjs.js",
```
should become
```
"type": "module",
"main": "dist/my-package.esm.js",
```

Conversely,
```
"main": "dist/my-package.esm.js",
```
should become
```
"main": "dist/my-package.cjs.js",
```
as it's not an ESM package.

If you fix this, not only will it make Preconstruct behave correctly, but it will also slash build times for packages with `"type": "module"` due to CommonJS bundles no longer being built for such packages.

tmm:
Adds support for ESM-only builds when [`"type": "module"`](https://nodejs.org/api/packages.html#type) is added to `package.json`.

- [x] Add support to init, validate, fix, dev, and build commands
- [x] Update and add tests
- [x] Update docs for `"type": "module"`
- [ ] Add changeset for release

Closes #471

Happy to chat about this more or modify the approach. Some open questions:

- [ ] Should this get added behind an experimental flag?
- [ ] What version bump does this change receive?


mitchellhamilton:
Hey, thanks for opening this PR

I’ve been thinking about what exactly Node ESM support should like in Preconstruct for a few years now and I’m still not sure what exactly the right answer is but I think it’d be worth trying something and to start with, I’m thinking “let’s only support all the new things and do things “correctly” and see how painful it is” would be an interesting starting place so I’ve pushed some changes to the documentation with the behavior I’m imagining. Let me know what you think. Apologies that this isn’t exactly a quick fix.

cc @Andarist curious if you have any thoughts on this

packages/cli/src/__tests__/fix.ts
```diff
@@ -982,3 +982,109 @@ test("project level exports field config", async () => {
...
+      type: "module",
+      preconstruct: {
+        entrypoints: ["index.js", "one.js", "two.js"],
+      },
+    }),
+    "src/index.js": "",
+    "src/one.js": "",
+    "src/two.js": "",
+    "one/package.json": JSON.stringify({}),
+    "two/package.json": JSON.stringify({}),
```
Andarist, line 1061: It's probably a safe bet to assume that if a tool understands `type: 'module'` then it also understands the `package.json#exports` field. So assuming that, we could stop generating nested `package.json` directories for entrypoints and rely on `package.json#exports` entirely for entrypoints.

packages/cli/src/build/config.ts
```diff
@@ -125,26 +130,27 @@ export function getRollupConfigs(pkg: Package, aliases: Aliases) {
...
-        chunkFileNames: "dist/[name]-[hash].cjs.prod.js",
-        dir: pkg.directory,
-        exports: "named",
-        interop,
-        plugins: cjsPlugins,
-      },
-    ],
-  });
+  if (!hasModuleType)
+    configs.push({
```
Andarist, line 134: Those files were also acting as optimizations for node since reading from `process.env` is slow. In ESM-only world, we should probably leverage development/production conditions in `package.json#exports` to apply the same optimization. I know that @mitchellhamilton had some concerns related to that though

packages/cli/src/dev.ts
```diff
@@ -184,44 +184,50 @@ export default async function dev(projectDir: string) {
...
+              )
+            );
+          }
 
-unregister();
-`
-            ),
-          ];
-          if (entrypoint.json.module) {
+          if (entrypoint.json.module || entrypoint.json.type === "module") {
```
Andarist, line 230: general question - should we even bother supporting "mixed" packages where some entrypoints are of type module and some are of type commonjs? IMHO the added complexity isn't quite worth it

packages/cli/src/messages.ts
```diff
@@ -22,7 +22,8 @@ export let errors = {
     `all entrypoints in a package must have the same fields and one entrypoint in this package has a ${field} field but you've declined the fix`,
   missingBrowserConditionWithFieldPresent: `the exports field is configured and the browser field exists in this package but it is not specified in the preconstruct.exports.envConditions field`,
   missingBrowserFieldWithConditionPresent: `the exports field is configured and the browser condition is set in preconstruct.exports.envConditions but the field is not present at the top-level`,
-  noModuleFieldWithExportsField: `when using the exports field, the module field must also be specified`,
+  noModuleFieldWithExportsField: `when using the exports field, the module field or type must also be specified`,
+  moduleFieldWithModuleType: `when \`"type": "module"\`, the module field is not allowed`,
```
Andarist, line 26:
```suggestion
  moduleFieldWithModuleType: `when \`"type": "module"\`, the module field should be redundant, and thus it is not allowed`,
```

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
 }
 ```
 
+### `type` (experimental) {#type}
+
+```ts
+"module" | "commonjs";
+```
+
+> If you're just thinking "I want to write code with the native ECMAScript import and export syntax and have my code work for most people", you likely should not use this feature. This feature allows you to build packages that import Node.js ESM only dependencies and therefore only support being imported from ESM in newer versions of Node.js and some bundlers.
```
Andarist, line 232: OTOH, it's not that uncommong to have self-contained repositories without external dependencies. In such it should be possible to use this feature without any "fear" and without deeper understanding of things. It's a fine line though and once you opt into this then you might end up with problems in the future, once you start adding external deps :shrug:

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
...
+
+The `type` config being set to `"module"` allows you to generate ECMAScript modules(ESM) compatible with Node.js' implementation of ESM (Node.js ESM).
```
Andarist, line 234:
```suggestion
The `type` config being set to `"module"` allows you to generate ECMAScript modules(ESM) compatible with Node.js' implementation of ESM (Node.js ESM), most modern bundlers, etc.
```

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
...
+
+You should likely only use this feature if you fit into one these categories:
+
+- You're interested in the state of modules in the JS ecosystem
+- You want to publish a package that synchronously imports a Node.js ESM only dependency
```
Andarist, line 239: "synchronously" might be a little bit misleading since it's not technically correct, imports are always fetched asynchronously - it's the initialization of the importing module that is deferred

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
...
+
+Node.js' implementation of ESM does not allow importing an ESM module from a CommonJS module synchronously. This means a given package only shipping Node.js ESM means that all of the packages that depend on it either need to use a dynamic `import()` call (which is asynchronous) to import it or only provide Node.js ESM and therefore enforce the same constraints on consumers.
+
+Preconstruct's support for Node.js ESM is currently an experiment to see how Preconstruct would work if it followed Node.js' behaviour for ESM. Note this behaviour means the resulting package will not support tools that do not support newer features like the `exports` field.
+
+- The [`exports` field feature](#exports) must also be enabled.
+- There are no entrypoint `package.json`s, the entrypoints are only specified in the `exports` field
+- The dist files for all entrypoints are in the `dist` directory in the root of the package
```
Andarist, line 247: This could be true regardless of this new option. Could we align the behavior for regular builds?

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
...
+- No `main`, `module`, `browser` or `umd:main` fields are used
```
Andarist, line 248: might be worth mentioning the available conditions here

site/docs/configuration.md
```diff
@@ -223,6 +223,45 @@ Preconstruct will enforce that the `exports` field that is written can is direct
...
+- Module resolution internally follows [Node.js ESM module resolution](https://nodejs.org/api/esm.html#import-specifiers) which requires explicitly writing extensions in imports. This includes following `"moduleResolution": "nodenext"` in TypeScript which means writing `.js` to import a TypeScript file.
+- The default export of a CommonJS dependency(even if it provides ESM intended for bundlers) will be the whole exports object, not `exports.default` if `exports.__esModule` is set and otherwise the whole exports object which is the default behaviour in Preconstruct.
```
Andarist, line 250: This could be rephrased somehow as it's hard to read and grok. I understand how it works but not sure if I would be able to fully understand it after reading just this point

Andarist:
Yea, I think that "use it at your own risk" is a good strategy and a way for us to test things out.

site/docs/configuration.md
Andarist: It's also not completely true that all deps have to be "esm only". It's very nuanced though and I'm not sure how to touch on that subject here, maybe we should link to a further explanation that could be put somewhere below? (you already touch the subject [here](https://github.com/preconstruct/preconstruct/pull/485/files/2666d1a3083b570b974f687c135717a7da160833#diff-0e368d7f259d5c89a12da2a63afa0229b7a43e1bf65148d85ed2fb34f6131398R250))

site/docs/configuration.md
mitchellhamilton: Yeah, I was thinking that as well

site/docs/configuration.md
mitchellhamilton:
```suggestion
- You want to publish a package that statically imports a Node.js ESM only dependency
```

tmm:
@mitchellhamilton @Andarist thanks for the thoughts! Cool if I make some updates that better match the updated docs and your feedback?

Andarist:
I would say - go for it 😉

mitchellhamilton:
yep!

mitchellhamilton:
Just want to note that https://github.com/microsoft/TypeScript/issues/50152 might affect what we want to do here

Andarist:
An explicit `types` condition for each package export should always work though, right? I don't imagine a world in which it wouldn't - that would be madness 🤣

mitchellhamilton:
Yeah, the design documented here would still work under `"moduleResolution": "node16"`. One part of that issue though is essentially saying "unless people are actually using Node for module resolution (which a lot of the time, they are not) `"moduleResolution": "node16"`" so Preconstruct building packages that only work with `"moduleResolution": "node16"` might not make sense.

packages/cli/src/build/config.ts
tmm: Good with adding this in. @mitchellhamilton any thoughts?

mitchellhamilton: I think it should be done as a separate change to this. The big thing IMO is that the syntax for it shouldn't be `process.env.NODE_ENV` since you could have `process.env.NODE_ENV === 'production'` but not have the `production` condition set.

PR #492, opened 2022-08-31
Upgrade Rollup, its plugins, Jest and more

Andarist:
For now, it's just a draft - because Rollup v3 has not been released yet as stable. It's just a PR to get ourselves ready for its release. Since Rollup v3 bumps the node version requirement - we probably should release a new major too if we want to land this right after their release.


packages/cli/src/build/__tests__/basic.ts
```diff
@@ -886,34 +886,34 @@ test("UMD with dep that uses process.env.NODE_ENV", async () => {
...
-          //# sourceMappingURL=scope-test.umd.min.js.map
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/scope-test.umd.min.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    !function(e,o){"object"==typeof exports&&"undefined"!=typeof module?module.exports=o():"function"==typeof define&&define.amd?define(o):(e="undefined"!=typeof globalThis?globalThis:e||self).test=o()}(this,(function(){"use strict";console.log("production");return"something"}));
+    //# sourceMappingURL=scope-test.umd.min.js.map
 
-          ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/scope-test.umd.min.js.map ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
-          {"version":3,"file":"scope-test.umd.min.js","sources":["../src/index.js"],"sourcesContent":["import { x } from \\"somewhere\\";\\nconsole.log(x);\\nexport default \\"something\\";"],"names":["console","log"],"mappings":"qOACAA,QAAQC,wBACO"}
-        `);
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/scope-test.umd.min.js.map ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    {"version":3,"file":"scope-test.umd.min.js","sources":["../src/index.js","../node_modules/somewhere/index.js"],"sourcesContent":["import { x } from \\"somewhere\\";\\nconsole.log(x);\\nexport default \\"something\\";","export let x = process.env.NODE_ENV;"],"names":["console","log"],"mappings":"qOACAA,QAAQC,ICD2B,oBDEpB"}
```
Andarist, line 915: I didn't notice this change previously, it seems like a possible improvement though 🤔

packages/cli/src/build/__tests__/basic.ts
```diff
@@ -1211,6 +1211,6 @@ test("importing css fails with a nice error", async () => {
     "src/blah.css": "",
   });
   await expect(build(dir)).rejects.toMatchInlineSnapshot(
-    `[Error: 🎁 @scope/test only .ts, .tsx, .js, .jsx, and .json files can be imported but "./blah.css" is imported in "src/index.js"]`
+    `[Error: 🎁 @scope/test only .ts, .tsx, .js, .jsx, and .json files can be imported but "blah.css" is imported in "src/index.js"]`
```
Andarist, line 1214: The change is caused by the added `path.relative` call here: https://github.com/preconstruct/preconstruct/pull/492/files#diff-81fa7127f1a6e4820814f89af91be081857155992595217ccf33830d6116e91eR59

Right now `source` is already resolved - I don't know how to trace it back to how it was written in the original source code. If we can't trace it back perhaps we should add `./` to computed relative paths if they happen to be in the same directory as the importer?

packages/cli/src/build/index.ts
```diff
@@ -48,9 +48,6 @@ function writeOutputFile(
           outputFile.map.toString()
         );
       }
-      if (outputOptions.sourcemap !== "hidden") {
```
Andarist, line 51: it seems that they now emit this on their own so we don't have to append it manually

packages/cli/src/rollup-plugins/flow-and-prod-dev-entry.ts
```diff
@@ -56,7 +56,10 @@ export default function flowAndNodeDevProdEntry(
         if (!resolved.external && !allowedExtensionRegex.test(resolved.id)) {
           warnings.push(
             new FatalError(
-              `only .ts, .tsx, .js, .jsx, and .json files can be imported but "${source}" is imported in ${
+              `only .ts, .tsx, .js, .jsx, and .json files can be imported but "${path.relative(
```
Andarist, line 59: both of those added `path.relative` calls could potentially be hoisted (the call is the same), maybe they even need to be wrapped with `normalizePath` but I'm not sure if that's required here yet. They might also contain a bug right now because they assume that `importer` is defined but according to the types (and some of the surrounding code here) it's actually nullable - the tests "pass" though and this doesn't crash. It would be good to add tests covering for the case when the `importer` is missing.

PR #495, opened 2022-09-23
fix: Pass tsconfig path to typescript parseJsonConfigFileContent

### LINKED ISSUE #443: The "fileName" or "name" properties of emitted files must be strings that are neither absolute nor relative paths, received "../../dist/index.d.ts".
Honestly, I have no idea what is going on here or how to fix it. I've found another issue on GitHub with the same error message but it looks like that was fixed & it was also a while ago: https://github.com/egoist/rollup-plugin-postcss/issues/250

This branch on the repo alii/nextkit seems to be running into the issue: https://github.com/alii/nextkit/tree/0b35d155e5902a9e76421b12da1a709479d4e9a7

It's probably something wrong in my setup but I've scoured through and can't seem to see anything out of the normal. Preconstruct is an amazing tool and has been flawless up until now 😧 thank you for the hard work!

image

mycroes:
Fix #443

### DIFF
--- a/.changeset/hip-flies-dance.md
+++ b/.changeset/hip-flies-dance.md
@@ -2,4 +2,4 @@
 "@preconstruct/cli": patch
 ---
 
-Pass config file path to typescript so relative paths are applied to correct directory
+The `outDir` and `declarationDir` tsconfig options are now ignored. These options are unnecessary for Preconstruct since it controls where the `.d.ts` files are emitted to. This fixes confusing errors if you had these options set.
--- a/.changeset/hip-flies-dance.md
+++ b/.changeset/hip-flies-dance.md
@@ -2,4 +2,4 @@
 "@preconstruct/cli": patch
 ---
 
-Pass package path to typescript so relative paths are applied to correct directory
+Pass config file path to typescript so relative paths are applied to correct directory
--- a/.changeset/hip-flies-dance.md
+++ b/.changeset/hip-flies-dance.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Pass package path to typescript so relative paths are applied to correct directory
--- a/packages/cli/src/build/__tests__/src-root.ts
+++ b/packages/cli/src/build/__tests__/src-root.ts
@@ -1,0 +1,83 @@
+import build from "../";
+import { testdir, getDist, ts, repoNodeModules } from "../../../test-utils";
+
+test("monorepo with root src dir and typescript package", async () => {
+  let tmpPath = await testdir({
+    "package.json": JSON.stringify({
+      name: "root",
+      devDependencies: {
+        typescript: "^3.8.3",
+      },
+      preconstruct: {
+        packages: ["src/packages/*"],
+      },
+    }),
+    ".babelrc": JSON.stringify({
+      presets: [
+        require.resolve("@babel/preset-env"),
+        require.resolve("@babel/preset-typescript"),
+      ],
+    }),
+    node_modules: {
+      kind: "symlink",
+      path: repoNodeModules,
+    },
+    "src/packages/my-package/package.json": JSON.stringify({
+      name: "my-package",
+      main: "dist/my-package.cjs.js",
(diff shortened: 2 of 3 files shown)


mycroes:
Had to alter memoize to take two arguments, but since it's only used for typescript memoization that didn't seem like much of an issue. I honestly doubt the need for the dirname in the memoization though, I don't think there's another path that will result in a different dirname/tsconfig combination.

mycroes:
I'm actually having slight second thoughts on this PR, as pointed out in #443 this problem occures if packages are below a `src/` directory in the workspace root. Since I replaced `process.cwd()` with `dirname` argument, is there any chance that it's not properly changing CWD if it detects a `src/` in the current directory? This change does mitigate the error, not sure if the root cause is properly handled though...

mitchellhamilton (commented):
The change itself seems fine but could you add some tests? I'm not sure this will actually fix the issue

mycroes:
I'll see what I can do. It's working, I'm using the changes using `yarn patch`, my repo is building fine now.

mycroes:
@mitchellhamilton I added a test which breaks without the path change and passes with the change. I hope this satisfies your requirements.

Also, when checking this out on Windows I get a ton of eslint errors due to line ending style, not sure if you have any advice on this. It's hugely annoying, hiding the actual style errors.

packages/cli/src/build/__tests__/src-root.ts
```diff
@@ -0,0 +1,83 @@
...
+    "src/packages/my-package/src/index.ts": ts`
+                                              export default {};
+                                            `,
+  });
+
+  await build(tmpPath);
+
+  expect(await getDist(tmpPath + "/src/packages/my-package"))
+    .toMatchInlineSnapshot(`
+  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/declarations/build/index.d.ts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```
mitchellhamilton, line 51: The fact that `outDir` is respected is quite bad imo since this doesn't really do what you might expect (and it would still break if for example you had a tsconfig in your project root with `outDir`) so I'd rather just ignore the option. Could you change this to delete `outDir` from the parsed config like how `declaration`, `emitDeclarationOnly` and `noEmit` are set? (Ideally, preconstruct would probably error if `outDir` is set but that would potentially be breaking so let's avoid that for now)

mycroes: Made the change as requested, also unset `declarationDir` because that has a similar effect.

mycroes:
I'm not sure about denying `outDir` or `declarationDir`, I actually have them set because in the same monorepo I also have a sample project that's built with Vite. I believe I can't have the packages set to `noEmit` when the Vite sample references them and otherwise I would either have to omit `tsc` to validate upfront or my source dir gets cluttered with the `tsc` build output (which now goes to a gitignored directory).

I don't know if preconstruct offers me the same functionality in dev mode, but this approach (with path aliases for local packages in sample tsconfig) gives me HMR on package changes as well, which makes for an extremely smooth development experience.

mycroes:
> Also, when checking this out on Windows I get a ton of eslint errors due to line ending style, not sure if you have any advice on this. It's hugely annoying, hiding the actual style errors.

I actually set `core.autocrlf` to false now, that at least preserves LF in the source files which makes the development experience way better.

.changeset/hip-flies-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Pass config file path to typescript so relative paths are applied to correct directory
```
mitchellhamilton, line 5:
```suggestion
The `outDir` and `declarationDir` tsconfig options are now ignored. These options are unnecessary for Preconstruct since it controls where the `.d.ts` files are emitted to. This fixes confusing errors if you had these options set.
```

mycroes:
You're welcome, thanks for preconstruct and emotion!

PR #516, opened 2023-02-09
Preserve `"use client"` module boundaries

emmatown:


packages/cli/src/rollup-plugins/server-components.ts
```diff
@@ -0,0 +1,75 @@
...
+export function serverComponentsPlugin({
+  sourceMap,
+}: {
+  sourceMap: boolean;
+}): Plugin {
+  return {
+    name: "server-components",
+
+    transform(code, id) {
+      if (code.includes("'use client'") || code.includes('"use client"')) {
```
Andarist, line 14:
```suggestion
      if (/['"]use client['"]/.test(code)) {
```

packages/cli/src/rollup-plugins/server-components.ts
```diff
@@ -0,0 +1,75 @@
...
+              ast: (ast as unknown) as AcornNode,
+            };
+
+            if (sourceMap) {
+              output.map = magicString.generateMap({ hires: true });
+            }
+            if (!output.meta) {
+              output.meta = {};
+            }
+            output.meta.isUseClientEntry = true;
```
Andarist, line 50: perhaps it would be a good idea to prefix this property with something like `_preconstruct_`?

packages/cli/src/rollup-plugins/server-components.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+      }
+      return null;
+    },
+    renderChunk(code, chunk) {
+      if (chunk.facadeModuleId !== null) {
+        const moduleInfo = this.getModuleInfo(chunk.facadeModuleId);
+        if (moduleInfo?.meta.isUseClientEntry) {
+          const magicString = new MagicString(code);
+          magicString.prepend('"use client";\n');
```
Andarist, line 62: i guess this is because Rollup might drop the directive, right?

packages/cli/src/rollup-plugins/server-components.ts
```diff
@@ -0,0 +1,75 @@
...
+          if (
+            node.type !== "ExpressionStatement" ||
+            node.expression.type !== "Literal" ||
+            typeof node.expression.value !== "string"
+          ) {
+            return;
+          }
+          if (node.expression.value === "use client") {
+            let magicString = new MagicString(code);
+            this.emitFile({ type: "chunk", id });
```
Andarist, line 33: I guess this is the core of the change that allows Rollup to preserve this file as a separate chunk, right?

packages/cli/src/build/__tests__/other.ts
```diff
@@ -824,3 +824,101 @@ test(".d.ts", async () => {
...
+    "src/c.js": js`
+                  import { D } from "./d";
+                  export function C() {
+                    return D;
+                  }
+                `,
+    "src/d.js": js`
+                  "use client";
+                  export const D = "d";
+                `,
```
Andarist, line 854: those end up being unused - was this intentional?

packages/cli/src/rollup-plugins/server-components.ts
emmatown: Rollup actually throws a warning if you have directives so this removes the directive and then adds it back.

packages/cli/src/rollup-plugins/server-components.ts
emmatown: We control all the Rollup plugins used so it doesn't really matter

packages/cli/src/rollup-plugins/server-components.ts
emmatown: Yes

packages/cli/src/build/__tests__/other.ts
emmatown: No, need to fix that, it was from debugging

packages/cli/src/rollup-plugins/server-components.ts
Andarist: Do u know if that's also the case in Rollup 3 (IIRC we are still on Rollup 2 here)?

packages/cli/src/rollup-plugins/server-components.ts
Andarist: Ah, right - good point. I forgot that this isn't anyhow exposed.

packages/cli/src/rollup-plugins/server-components.ts
emmatown: It is - https://github.com/rollup/rollup/blob/e0d9681de58cee80f162d02c65de293954c4a13f/src/ast/nodes/ExpressionStatement.ts#L20

Andarist: Interesting, it might be worth creating a Rollup issue about this. I suspect that more people might need to handle those directives in the near-ish future.

emmatown: tbh, imo the current behaviour is the most sensible option. There's no correct answer to bundling directives so erroring on it seems like the ideal thing.

PR #536, opened 2023-04-04
Upgrade dependency `terser`

jordanoverbye:
This PR upgrades the dependency on `terser` from the `cli` package.

 The current version has a [security warning (CVE-2022-25858)](https://github.com/advisories/GHSA-4wf5-vphf-c2xc).

For reference, here is a link to the [Terser changelog ](https://github.com/terser/terser/blob/master/CHANGELOG.md).


emmatown (approved):
The newer version would already be used when people install Preconstruct and this isn't really a vulnerability but sure, updating the requirement is fine (Preconstruct executes arbitrary code by loading Babel configs so a ReDoS is meaningless)

PR #542, opened 2023-04-13
Introduce new `useDevProdConditions` config

Andarist:


packages/cli/src/build/__tests__/build.ts
```diff
@@ -907,3 +907,102 @@ test("self import", async () => {
...
+    };
+
+    export { thing };
+
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/dev-prod-conditions-test.cjs.dev.js ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    'use strict';
+
+    Object.defineProperty(exports, '__esModule', { value: true });
+
+    const thing = process.env.NODE_ENV !== "production" ? {
```
Andarist, line 979: note to myself: this one lacks the replacing output plugin. However, most likely I need to "fork" using input plugins and I'm somewhat surprised that this `.cjs.dev.js` bundle isn't already replaced using one. I guess this might allow for dynamic~ runtime-based switching but I never needed that and I think that with `exports` it's not needed.

packages/cli/src/build/config.ts
```diff
@@ -198,14 +220,19 @@ export function getRollupConfigs(pkg: Package, aliases: Aliases) {
...
         },
-        hasModuleField && {
-          format: "es" as const,
-          entryFileNames: `[name].${getDistExtension("browser-esm")}`,
-          chunkFileNames: `dist/[name]-[hash].${getDistExtension(
-            "browser-esm"
-          )}`,
-          dir: pkg.directory,
-        },
+        ...(hasModuleField
```
Andarist, line 223: I'm not sure why `browser` configs depend on `hasModuleField` here but perhaps it's always validated to be true when `exports` are enabled?

packages/cli/src/build/config.ts
```diff
@@ -198,14 +220,19 @@ export function getRollupConfigs(pkg: Package, aliases: Aliases) {
...
+          ? forkDevProdConditions(
```
Andarist, line 224: I feel like it's not bad to "fork" those things here like that - for the browser and worker bundles. Both prod and dev bundles are colocated etc. The node thing though... that's a little bit of a dance. It took me a while to understand how things around it works. It's just a thought but maybe this could be refactored somehow in the future.

packages/cli/src/build/config.ts
```diff
@@ -66,7 +67,29 @@ const babelHelperId = /@babel\/runtime(|-corejs[23])\/helpers\//;
...
+  }
+  return (["dev", "prod"] as const).map((condition) => ({
+    ...options,
+    entryFileNames: forkExtension(options.entryFileNames, condition),
+    chunkFileNames: forkExtension(options.chunkFileNames, condition),
+    plugins: [
+      inlineProcessEnvNodeEnv({
+        sourceMap: false,
+        value: condition === "prod" ? "production" : "development",
+      }),
```
Andarist, line 85: I recall we were discussing in the past using something else than `process.env.NODE_ENV` for this. It's a somewhat weird-ish convention that abuses the `process.env` in a way that it was not designed to be abused :P What are your thoughts about this? Do you want to continue using this? Or would you like to switch to some `import.meta.preconstruct.isDevelopment`?

packages/cli/src/utils.ts
```diff
@@ -86,25 +144,14 @@ export function exportsField(
...
-        : esmBuild,
-      default: getExportsFieldOutputPath(entrypoint, "cjs"),
-    };
-
     output[
       "." + entrypoint.name.replace(entrypoint.package.name, "")
-    ] = exportConditions;
+    ] = exportsFieldConfig.useDevProdConditions
+      ? {
+          development: entryPointExports(entrypoint, exportsFieldConfig, "dev"),
```
Andarist, line 151: what should be done about TS declarations for this? I think it's OK to not worry about this **now** but ideally it should also work with `tsconfig.json#compilerOptions.customConditions: ['development']`

emmatown:
So I've actually been thinking about this problem space recently and I've been planning to solve it in a different way (including replacing `exports.envConditions`)

Basically, you can add to the `imports` field, Preconstruct will calculate all the combinations of conditions that could exist and generate the bundles necessary. We would ban using the `import`, `require`, `module`, and `types` conditions in `imports`, require a `default` for everything (we would use the `default` for types, maybe eventually we would do "you can have different types for different conditions" but for now, no), ban conditions that would never match and enforce consistent ordering of conditions so we can actually generate a correct `exports` field.

So for example for dev/prod, you could add this to your `package.json`:

```json
"imports": {
  "#development": {
    "development": "./src/development.ts",
    "default": "./src/production.ts"
  }
}
```

And add files like this:

```ts
// src/development.ts
export const isDev = true
// src/production.ts
export const isDev = false
```


and you import from `#development` and use it however you want, Preconstruct would figure out the right `exports` conditions to have and generate all the bundles.

(I've checked and Rollup can handle dead-code elimination with exports like this)

It's nice in a bunch of ways:

- It's not some special Preconstruct thing, so you don't have to learn some new API and it'll work nicely with `preconstruct dev`
- Preconstruct doesn't have to care about any of conditions, so you can use whatever other conditions
- You can replace whole files so you can fork importing and etc.
- We don't have to do replacing magic things like `process.env.NODE_ENV` or `typeof window`
- You could type-check for each fork with the `customConditions` thing in tsc and it would just work.

Also, not something I'd like to do immediately but for after the initial version of this: a config like `preconstruct.exports.staticConditions` where instead of generating those conditions into the `exports` field, you would provide whichever of the static conditions to `preconstruct build` that you want to build with so you can have build-time feature flags so you could e.g. have an `experimental` condition which adds some new features and you can publish the stable thing and the experimental thing separately. And you can still have nice type checking for with `customConditions`.

Andarist:
All of that sounds nice - especially for **custom** conditions. I feel though there are some huge DX problems around semi-standard conditions:
- maintaining the list of imports/conditions is a chore
- generating extra files as well
- TS might have problems when it comes to import completions for `imports`, although maybe that's just for subpath imports
- because it relies on the file system and stuff, there is a disconnect between an attempt to use a condition and generating the appropriate file. Similarly there is a disconnect between removing the usage of a certain condition and removing its related "configuration" (and files on disk)

Perhaps a package like `@preconstruct/conditions` could address most of those concerns? It's still not idea that one would have to actually install it to start using this.

emmatown:
I mostly agree with your concerns though I think any improvements to those can be layered on so I'm going to do the initial version with just the necessary stuff and we can layer on usability improvements/syntax sugar in the future

emmatown:
> generating extra files as well

With this, are you referring to having the `development.ts` and `production.ts` files? or something else?

Andarist:
> I mostly agree with your concerns though I think any improvements to those can be layered on so I'm going to do the initial version with just the necessary stuff and we can layer on usability improvements/syntax sugar in the future

Sounds like a plan. Let me know if you want any help with this but at the moment it feels like you have a better grasp of the solution that you want here.

> With this, are you referring to having the development.ts and production.ts files? or something else?

This.

Andarist:
superseded by https://github.com/preconstruct/preconstruct/pull/551

PR #543, opened 2023-04-14
Fixed the default export shape in strict ESM environments

Andarist:
The time has come, brace yourself, yadayada 😉

### DIFF
--- a/packages/cli/src/build/__tests__/build.ts
+++ b/packages/cli/src/build/__tests__/build.ts
@@ -936,7 +936,7 @@ test("correct default export using mjs and dmts proxies", async () => {
       preconstruct: {
         entrypoints: ["index.ts", "something.ts"],
         exports: {
-          unwrappedDefaultExportForImportCondition: true,
+          importDefaultExport: "unwrapped-default",
         },
       },
     }),
@@ -1158,7 +1158,7 @@ test("no __esModule when reexporting namespace with mjs proxy", async () => {
       preconstruct: {
         entrypoints: ["index.js"],
         exports: {
-          unwrappedDefaultExportForImportCondition: true,
+          importDefaultExport: "unwrapped-default",
         },
       },
     }),
--- a/packages/cli/src/build/__tests__/build.ts
+++ b/packages/cli/src/build/__tests__/build.ts
@@ -735,6 +735,33 @@ test("worker and browser build", async () => {
   `);
 });
 
+const tsSetupFiles = {
+  "node_modules/typescript": {
+    kind: "symlink",
+    path: path.join(repoNodeModules, "typescript"),
+  },
+  "node_modules/@babel/preset-typescript": {
+    kind: "symlink",
+    path: path.join(repoNodeModules, "@babel/preset-typescript"),
+  },
+  "babel.config.json": JSON.stringify({
+    presets: ["@babel/preset-typescript"],
+  }),
+  "tsconfig.json": JSON.stringify({
+    compilerOptions: {
+      module: "NodeNext",
+      moduleResolution: "nodenext",
+      strict: true,
+      declaration: true,
+    },
+  }),
+} as const;
+
+const tscBinPath = path.join(
+  path.dirname(require.resolve("typescript/package.json")),
+  "bin/tsc"
+);
+
 test("typescript with nodenext module resolution", async () => {
   let dir = await testdir({
     "package.json": JSON.stringify({
@@ -776,11 +803,6 @@ test("typescript with nodenext module resolution", async () => {
     "packages/pkg-a/not-exported.ts": ts`
       export const notExported = true;
     `,
-
-    "packages/pkg-a/node_modules": {
-      kind: "symlink",
-      path: repoNodeModules,
-    },
     "blah.ts": ts`
       import { thing } from "pkg-a";
       import { something } from "pkg-a/something";
@@ -795,28 +817,14 @@ test("typescript with nodenext module resolution", async () => {
       acceptThing<"other">(thing); // should error
       acceptThing<"other">(something); // should error
     `,
-    "tsconfig.json": JSON.stringify({
-      compilerOptions: {
-        module: "NodeNext",
-        moduleResolution: "nodenext",
-        strict: true,
-        declaration: true,
-      },
-    }),
+    ...tsSetupFiles,
   });
   await fs.ensureSymlink(
     path.join(dir, "packages/pkg-a"),
     path.join(dir, "node_modules/pkg-a")
   );
   await build(dir);
-  let { code, stdout, stderr } = await spawn(
-    path.join(
-      path.dirname(require.resolve("typescript/package.json")),
-      "bin/tsc"
-    ),
-    [],
-    { cwd: dir }
-  );
+  let { code, stdout, stderr } = await spawn(tscBinPath, [], { cwd: dir });
   expect(code).toBe(2);
   expect(stdout.toString("utf8")).toMatchInlineSnapshot(`
     "blah.ts(3,29): error TS2307: Cannot find module 'pkg-a/not-exported' or its corresponding type declarations.
@@ -957,11 +965,6 @@ test("correct default export using mjs and dmts proxies", async () => {
       export const notExported = true;
       export default "foo";
     `,
-
-    "packages/pkg-a/node_modules": {
-      kind: "symlink",
-      path: repoNodeModules,
-    },
     "blah.mts": ts`
       function acceptThing<T>(x: T) {}
 
@@ -1022,14 +1025,7 @@ test("correct default export using mjs and dmts proxies", async () => {
       acceptThing(indexNs.default, true);
       acceptThing(somethingNs.default, 100);
     `,
-    "tsconfig.json": JSON.stringify({
-      compilerOptions: {
-        module: "NodeNext",
-        moduleResolution: "nodenext",
-        strict: true,
-        declaration: true,
-      },
-    }),
+    ...tsSetupFiles,
   });
   await fs.ensureSymlink(
     path.join(dir, "packages/pkg-a"),
@@ -1049,9 +1045,7 @@ test("correct default export using mjs and dmts proxies", async () => {
     export default _default;
 
     ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ packages/pkg-a/dist/pkg-a.cjs.d.mts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
-    export {
-      thing
-    } from "./declarations/src/index.js";
+    export * from "./declarations/src/index.js";
     import ns from "./declarations/src/index.js";
     export default ns.default;
     //# sourceMappingURL=pkg-a.cjs.d.mts.map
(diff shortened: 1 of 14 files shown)


packages/cli/src/package.ts
```diff
@@ -350,6 +350,7 @@ type CanonicalExportsFieldConfig =
   | {
       envConditions: Set<"worker" | "browser">;
       extra: Record<string, JSONValue>;
+      useMjsProxy: boolean;
```
Andarist, line 353: quite frankly... I would just make this the default and remove the config. I introduced this with a config to limit the scope of the change - both for consumers (as you might want to preserve the existing behavior) and for the test suite (I didn't want to update existing tests just yet). 

I think that the previous behavior was already *broken* and I don't think of it as of a correct behavior - so I would personally treat this as a bug fix. It's subjective though 🤷‍♂️

packages/cli/src/utils.ts
```diff
@@ -197,6 +203,41 @@ export function tsTemplate(
...
+  const escapedPath = JSON.stringify(relativePath);
+  return `import * as _ns from ${escapedPath};\n${exports
+    .map((name, i) => {
+      if (name === "default") {
+        return `export default _ns.default.default;`;
+      }
+      if (!isValidJsIdentifier(name)) {
+        const escapedName = JSON.stringify(name);
+        return `var _export${i} = _ns[${escapedName}];\nexport { _export${i} as ${escapedName} };`;
+      }
```
Andarist, line 220: Rollup doesn't actually handle this today too well either, I opened a fix for this [here](https://github.com/rollup/rollup/pull/4939)

packages/cli/src/utils.ts
```diff
@@ -197,6 +203,41 @@ export function tsTemplate(
...
+export function dmtsTemplate(exports: string[], relativePath: string) {
+  const escapedPath = JSON.stringify(`${relativePath}.js`);
+  return `import * as _ns from ${escapedPath};\n${exports
+    .map((name, i) => {
+      if (name === "default") {
+        return `declare const _def: typeof _ns.default.default;\nexport default _def;`;
+      }
+      if (!isValidJsIdentifier(name)) {
+        throw new Error("TypeScript does not support non-identifier exports");
+      }
```
Andarist, line 235: I couldn't make it work with TS, it doesn't like this valid JS:
```js
export let a = 1;
export { a as "test" }
// Identifier expected.ts(1003)
// Cannot find module 'test' or its corresponding type declarations.ts(2307)
```

packages/cli/src/utils.ts
```diff
@@ -197,6 +203,41 @@ export function tsTemplate(
...
+        return `var _export${i} = _ns[${escapedName}];\nexport { _export${i} as ${escapedName} };`;
+      }
+      return `export var ${name} = _ns.${name};`;
+    })
+    .join("\n")}\n`;
+}
+
+export function dmtsTemplate(exports: string[], relativePath: string) {
+  const escapedPath = JSON.stringify(`${relativePath}.js`);
+  return `import * as _ns from ${escapedPath};\n${exports
```
emmatown, line 228: Can't this be an `export * from`? and then you'd just have the special handling for default? And I wouldn't expect this to work as written since the default isn't in `import/export *`?

packages/cli/src/utils.ts
emmatown: Again, similar to the types, this wouldn't work as written since the `default` isn't in the `import * as`. This should just be `export *` + stuff for the default.

packages/cli/src/package.ts
emmatown: Yeah, I'm planning to do a major soon to enable `exports` by default and the Rollup 3 upgrade and we can change the default for this in that. I'd probably want to keep the config though

packages/cli/src/build/__tests__/build.ts
```diff
@@ -907,3 +907,227 @@ test("self import", async () => {
...
+      },
+    }),
+    "packages/pkg-a/package.json": JSON.stringify({
+      name: "pkg-a",
+      main: "dist/pkg-a.cjs.js",
+      module: "dist/pkg-a.esm.js",
+      exports: {
+        ".": {
+          module: "./dist/pkg-a.esm.js",
+          import: "./dist/pkg-a.cjs.mjs",
```
emmatown, line 926: `.cjs.mjs` it's totally accurate but also 😂

packages/cli/src/utils.ts
Andarist: > And I wouldn't expect this to work as written since the default isn't in import/export *?

I also have thought that but it is present in `_ns`. I was testing this with bare node and also looked through transpiled interop helpers of Babel and TS. Perhaps this is a recent change to interop stuff or something?

Also note that I've added actual tests for this - both for `tsc` and `node` and it seems to work quite good right now. Let me know if you think that they are doing an incorrect thing or smth.

> Can't this be an export * from? 

Hm, yee. It seems that despite what I have just mentioned, this can be simplified like that. `_ns.default` is available in this mjs-proxy but it won't be available in the consuming code:

When I import from this "mjs proxy" file:
```js
export * from './cjs-lib.cjs'
import * as nsProxy from './cjs-lib.cjs'

console.log('nsProxy', nsProxy)
```

we get those logs:
```
nsProxy [Module: null prototype] {
  __esModule: true,
  default: { foo: 1, default: "i'm here" },
  foo: 1
}
```

packages/cli/src/build/__tests__/build.ts
Andarist: ye, i actually meant to mention this weirdo in a comment but forgot. I came to the same conclusion - it looks weird as hell but it's kinda OK. Of course, it also doesn't really matter but at the same time u can come up with a reasoning why it is like this :p

Andarist: We could also just completely avoid using `.mjs` and nest this file in a directory with `package.json: {type:'module'}` but I'm not sure if it's worth the added complexity 🤷

emmatown: Oh no, I think this is the right choice, it's just funny

packages/cli/src/package.ts
Andarist: Fine by me 👍

packages/cli/src/utils.ts
Andarist: hm, we don't 🤔 this is in the `!isValidJsIdentifier` block. However, like you suggested - most of it can probably be simplified with `export *`. Unless you think now there is some reason to keep the current "explicit" strategy?

packages/cli/src/utils.ts
emmatown: Could we just emit the string for invalid identifiers so that when TS adds support, it'll just work?

packages/cli/src/utils.ts
emmatown: Oh sorry, I missed the `if`

packages/cli/src/package.ts
emmatown: Also, on naming, how about `unwrappedDefaultExportInStrictESM` or something like that to indicate what it does rather than how it does it?

Andarist: changed the name to the suggested one

packages/cli/src/package.ts
```diff
@@ -420,6 +422,15 @@ function parseExportsFieldConfig(
           name
         );
       }
+    } else if (key === "unwrappedDefaultExportForImportCondition") {
```
Andarist, line 425: I'd love to use this repo-wide, see a similar concern+question [here](https://github.com/preconstruct/preconstruct/pull/542#discussion_r1165762092)

packages/cli/src/utils.ts
Andarist: It's no longer a concern since we just emit `export * from`

emmatown (commented):
The proxy files need to be written for `preconstruct dev` as well (let’s just do an `export * from` there)

Andarist:
what about declaration maps and other source map things? i'm not sure how to handle those correctly

packages/cli/src/package.ts
emmatown: Yes to making it project-wide, an empty object should be valid and equivalent to true (i would imagine an empty object already works on a package)

emmatown:
We only generate js source maps for UMD builds so that isn’t a thing. For declaration maps, you should be able to just write the same declaration map we use for .cjs.d.ts but as .cjs.d.mts

emmatown:
Thinking about this more and the fact that I think the right answer is just to avoid default exports, I’d like to have three modes:
- the current "you need to do .default" thing
- what this PR implements with an import condition
- no import condition but fail builds which have default exports in entrypoints

For the config, probably this:

```ts
"importConditionDefaultExport": "namespace" |  "default" | "namespace-disallow-default"
```

It would be outside of the `exports` config. If you set it to `default` when `exports` isn't enabled, that would be an error. For now, it would default to `namespace`. In the next major, it would default to `namespace-disallow-default`.

Andarist:
While there are good reasons to avoid `default` exports, I don't think that it should be "mandated" by a tool. Different people like different things and `default` might even sometimes be **required** by some frameworks for different purposes.

I also believe that the problem is not even exclusive to packages with `default` export. Even without `default` this is still allowed:
```ts
import ns from 'cjs-lib'
```
I think that's not a good thing because it can accidentally be used by users and it's likely not how the lib's usage is documented.

emmatown:
Yeah, fair. I was initially thinking who cares about the namespace being technically available but i suppose if more people actually use native ESM, it'll be kinda annoying so sure.

I do think I prefer the config with string values instead of booleans though, so if you could make the config this (still inside of exports unlike my last comment suggests though + allow it to be specified at the project level), that would be great.
```ts
"importConditionDefaultExport": "namespace" |  "default"
```

packages/cli/src/build/__tests__/build.ts
```diff
@@ -1135,6 +1139,378 @@ test("correct default export using mjs and dmts proxies", async () => {
...
+  await fs.ensureSymlink(
+    path.join(dir, "packages/pkg-b"),
+    path.join(dir, "node_modules/pkg-b")
+  );
+  const error = await build(dir).catch((x) => x);
+  expect(stripAnsi(error + "")).toMatchInlineSnapshot(`
+    "Error: 🎁 Generating TypeScript declarations for packages/pkg-b/src/index.ts failed:
+    🎁 packages/pkg-b/src/index.ts:1:14 - error TS2841: The type of this expression cannot be named without a 'resolution-mode' assertion, which is an unstable feature. Use nightly TypeScript to silence this error. Try updating with 'npm install -D typescript@next'.
+    🎁
+    🎁 1 export const thing = import("pkg-a");
```
Andarist, line 1378: is this because TS already supported `import` in declaration files, it had its own semantics and stuff so now it can't use that syntax alone to differentiate between loading a thing with import/require semantics? 🤔

emmatown: yes

emmatown:
Re naming bike shedding: curious to hear reasoning behind the naming you've done?

I've added some docs using the naming I suggested. To me, `importConditionDefaultExport` makes more sense than `importDefaultExport` since it's specifically about the `import` condition and it doesn't affect importing the default export with ESM syntax in files that are not `type: module`/`.mjs`/etc. And I don't really see what the `unwrapped-` in `unwrapped-default` adds given the alternative value is `namespace`.

PR #546, opened 2023-05-02
Fix `importConditionDefaultExport`

emmatown:
The previous version didn't work with `"moduleResolution": "bundler"`, this fixes that.

### DIFF
--- a/packages/cli/src/dev.ts
+++ b/packages/cli/src/dev.ts
@@ -10,6 +10,10 @@ import {
   dmtsTemplate,
   mjsTemplate,
   getBaseDistFilename,
+  getJsDefaultForMjsFilepath,
+  jsDefaultForMjsTemplate,
+  dtsDefaultForDmtsTemplate,
+  getDtsDefaultForMtsFilepath,
 } from "./utils";
 import * as babel from "@babel/core";
 import * as fs from "fs-extra";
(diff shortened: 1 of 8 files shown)


packages/cli/src/dev.ts
```diff
@@ -266,24 +274,38 @@ unregister();
...
+                return Promise.all([
+                  fs.writeFile(
+                    filepath,
+                    mjsTemplate(
+                      // the * won't really do anything right now
+                      // since cjs-module-lexer won't find anything
+                      // but that could be fixed by adding fake things
+                      // to the .cjs.js file that look like exports to cjs-module-lexer
+                      // but don't actually add the exports at runtime like esbuild does
+                      // (it would require re-running dev when adding new named exports)
```
Andarist, line 295: I don't understand what's happening here but it sounds like it does nothing *today* so that's... fine? 😅

Andarist (approved):
If I understand correctly, the core of this PR is to avoid loading CJS file directly into MJS because TS doesn't understand `module` but yet it understands `import` **and** changes what kind of a default we get there (tangent... is this even OK? i guess that it might be because it kinda acts as `module` here even though not through that condition, OTOH... this isn't how strict bundlers work, I think). To avoid this we create an extra CJS file to normalize the default thing so it can mean the same thing for the runtime and types in more scenarios

packages/cli/src/dev.ts
emmatown: correct (this bit is also not new to this PR, it's just moved around)

The comment is generally about making this work in `preconstruct dev`:

```
// somewhere.mjs
import { something } from 'my-pkg'
```

It currently doesn't (and didn't before any of the `importConditionDefaultExport ` stuff).

PR #551, opened 2023-05-12
Imports conditions

emmatown:
This implements the design from https://github.com/preconstruct/preconstruct/pull/542#issuecomment-1508333012 under an experimental flag `importsConditions`. There's still some stuff I probably want to change here (hence why it's under an experimental flag) but I'd like to try it out and see how it is to use first.

A thing I didn't mention in the PR comment linked above that I've done here is for `imports` fields like this:

```json
{
  "imports": {
    "#a": {
      "worker": "./src/a.js",
      "browser": "./src/b.js",
      "default": "./src/a.js"
    }
  }
}
```

We'll only generate two bundles for each format rather than three since the worker and default conditions resolve to the same place. (we use the shortest combination of conditions to name bundles)

Also, in the generated `.d.ts` files, we re-write imports from `#something` to the resolved thing with the default condition applied. This is so that the `.d.ts` files work in `moduleResolution: node` (though of course, when authoring packages, you need to use bundler or nodenext so that `#something` can be resolved).

In the future, I'm thinking of having essentially two `.d.ts` outputs, one with `#something` replaced and one where it's not replaced, this would allow having different types for different conditions. I don't think that by itself would be that useful tbh since I'm guessing that most package consumers will not be configuring `customConditions` in TS. Though what would be quite useful is the `types@semver-range-of-ts-version` magical condition thing, specifically, we could essentially polyfill it for `moduleResolution: node` via `typesVersions` without users having to deal with `typesVersions`.  One slight problem with this is that without `isolatedDeclarations`, you could accidentally rely on type information from a particular implementation of `#something` and therefore generate the wrong `.d.ts` files so I'm thinking if/when we implement this part, it'll probably stay experimental until `isolatedDeclarations` exists. (we could of course also just generate `.d.ts` files for each set of conditions but that would be unacceptably slow imo)

Also, one random thing that's interesting with this design (though I don't think I'd say it's really a problem) is that you can conditionally have different exports for different conditions except that you couldn't conditionally have a default export. (the _implementations_ of a default export could be different for different conditions but you couldn't have a default export under some conditions and not under others)

cc @Andarist

Closes #542


packages/cli/src/utils.ts
```diff
@@ -84,37 +89,97 @@ export function exportsField(
...
     };
 
-    output[
-      "." + entrypoint.name.replace(entrypoint.package.name, "")
-    ] = exportConditions;
-  });
-  return {
-    ...output,
-    "./package.json": "./package.json",
-    ...exportsFieldConfig.extra,
```
Andarist, line 116: what has happened to those in the legacy mode? I mean, likely this works OK since tests pass but i wonder implementation-wise

PR #553, opened 2023-05-17
Use local TS version in VS Code

Andarist:


emmatown (commented):
Why?

Andarist:
Cause otherwise VS Code chooses to use the TS version that comes bundled in it - that may differ from the one that is installed locally and might create weird observable results. Just lately bumped into this in `changesets/action` since the local version was ancient (some `3.x`) - in the IDE everything looked OK but `tsc` was failing. IMHO, it's just way better to use a consistent TS version, and VS Code default behavior baffles me.

emmatown:
https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript seems to imply that the option here does nothing without going into the TypeScript version picker in VS Code and choosing the right thing (which you can already do without specifying this option) or additionally adding `"typescript.enablePromptUseWorkspaceTsdk": true` and accepting the prompt?

Andarist:
Hm, since `typescript.selectTypeScriptVersion` command **creates** this file after choosing "Use Workspace Version", I always assumed that this setting makes this consistent for all users of the workspace.

According to the quoted docs... it does not, lol. I understand that one could point to a different location with `typescript.tsdk` - and that's useful on its own (for example, TS codebase itself uses this to point to its local/dev version). But since the version in `node_modules/typescript` is discoverable by VS Code even before adding this setting... this is somewhat confusing (at least for me, since I operated on wrong conclusions about this **for years** 🤣 ). I created this file by executing that command.

In the VS Code's tooltip we can also read this:
```ts
- When specified as a user setting, the TypeScript version from typescript.tsdk automatically replaces the built-in TypeScript version.
- When specified as a workspace setting, typescript.tsdk allows you to switch to use that workspace version of TypeScript for IntelliSense with the TypeScript: Select TypeScript version command.
```

Kinda doesn't make sense to me but 🤷 Neither of those things is what I really want for my projects.

Thanks for mentioning `typescript.enablePromptUseWorkspaceTsdk`. it seems to be the best option out there to achieve my goals and I will start using it in my projects. Do you have any opinion about using this here? I already configured my VS Code to use the workspace version of TS so I'm good to go. However, ideally we'd either commit this together with `typescript.enablePromptUseWorkspaceTsdk` or add `.vscode/settings.json` to the `.gitignore`.

emmatown:
> add .vscode/settings.json to the .gitignore

Let's do this

Andarist:
Superseded by https://github.com/preconstruct/preconstruct/pull/557

PR #556, opened 2023-05-18
Replace TS extensions in generated declaration files

Andarist:


packages/cli/src/build/__tests__/build.ts
```diff
@@ -2191,3 +2191,54 @@ test("correct default export using mjs and dmts proxies with moduleResolution: b
   `);
   expect(node.stderr.toString("utf8")).toMatchInlineSnapshot(`""`);
 });
+
+test("replaces ts extensions in module specifiers within generated declarations", async () => {
```
Andarist, line 2195: i'll add more tests after we agree on the exact implementation

packages/cli/src/rollup-plugins/typescript-declarations/common.ts
```diff
@@ -195,11 +195,32 @@ export const getDeclarationsForFile = async (
...
+                }[match]!;
+              }
+            );
+          };
+          const cachedVisitModuleSpecifier = memoize(
+            visitModuleSpecifier
+              ? (moduleSpecifier) =>
+                  replaceTsExtensions(visitModuleSpecifier(moduleSpecifier))
+              : replaceTsExtensions
+          );
```
Andarist, line 223: Since `allowImportingTsExtensions: true` is allowed with just any `moduleResolution` type ( https://github.com/microsoft/TypeScript/pull/52230 ), I think it's the best/easiest to just apply this logic at all times. 

We could cross-reference if that option is enabled but it seems like an overcomplication to me. If it's not enabled then typechecking process would complain about it and it doesn't quite matter if we replace those or not when generating declaration files

packages/cli/src/rollup-plugins/typescript-declarations/common.ts
```diff
@@ -195,11 +195,32 @@ export const getDeclarationsForFile = async (
...
+              return moduleSpecifier;
+            }
+            return moduleSpecifier.replace(
+              /(\.d)?\.(ts|tsx|mts|cts)$/,
+              (match) => {
+                return {
+                  ".ts": ".js",
+                  ".tsx": ".js",
+                  ".mts": ".mjs",
+                  ".cts": ".cjs",
```
Andarist, line 209: I'm not quite sure how this would behave in older TS versions. It feels like a safe bet that those cases are not that popular today anyway and that we might triage the potential issues if somebody raises them. It also seems quite fair to say that using those extensions might require TS 4.8+ (or something like that) for package **consumers**.

packages/cli/src/rollup-plugins/typescript-declarations/common.ts
```diff
@@ -195,11 +195,32 @@ export const getDeclarationsForFile = async (
...
-          }
-
-          const cachedVisitModuleSpecifier = memoize(visitModuleSpecifier);
+          const replaceTsExtensions = (moduleSpecifier: string) => {
+            if (!moduleSpecifier.startsWith(".")) {
+              return moduleSpecifier;
+            }
+            return moduleSpecifier.replace(
+              /(\.d)?\.(ts|tsx|mts|cts)$/,
+              (match) => {
```
Andarist, line 204: I also decided to replace those with runtime equivalents because that works in older TS versions and it seems to be more compatible with the stricter rules imposed by `type:module` and stuff. While we don't really support `type:module` yet, we will at some point. So to reduce conditional logic for the future, I just went with replacing - instead of stripping those away.

emmatown (commented):
Two things:
- The implementation here would break if I have a file named `something.ts.ts` and import it with `./something.ts`
- I'd like to only have this behaviour under the `importsCondition` flag for now since that's the only place we currently replace module specifiers in TypeScript declarations

So with those two things in mind, the implementation should really just be delete these three lines:
https://github.com/preconstruct/preconstruct/blob/1147fee9cf6ee0a4d1174114c979f59bb531eb93/packages/cli/src/rollup-plugins/typescript-declarations/get-used-declaration-with-replacing.ts#L48-L50

Andarist:
> The implementation here would break if I have a file named something.ts.ts and import it with ./something.ts

I guess, I could try to check if sources for those things exist and only replace them conditionally. I'll look into it - I rechecked that this is allowed by TS and it doesn't trip over this in its module specifier validation logic etc.

> I'd like to only have this behaviour under the importsCondition flag for now since that's the only place we currently replace module specifiers in TypeScript declarations

If we are worried about some unintended consequences related to this (although... it might be hard to verify if there are any even after shipping if we can't figure out them now), I think this should be a separate experimental flag.

The behavior itself is not related anyhow to `importConditions` really and one can use `allowImportingTsExtensions: true`

> So with those two things in mind, the implementation should really just be delete these three lines:

Nice, I verified that it works so I could reuse that logic there (or rely on it if we'll do it only in combination with `importConditions`)

emmatown:
> If we are worried about some unintended consequences related to this (although... it might be hard to verify if there are any even after shipping if we can't figure out them now), I think this should be a separate experimental flag.

Yeah, this is what I'm worried about, wdyt if we do it in both `importsConditions` and `onlyEmitUsedTypeScriptDeclarations`?

Andarist:
How is that better than using a different experimental flag? Unless you treat those two modes as things that will become non-experimental soon-ish and as such this replacing logic will also become enabled *at all times* soon-ish.

emmatown:
> Unless you treat those two modes as things that will become non-experimental soon-ish and as such this replacing logic will also become enabled at all times soon-ish.

Yeah, I'm thinking `onlyEmitUsedTypeScriptDeclarations` will be the default soon. (`importsConditions` will probably take longer to get right)

Andarist:
@emmatown pushed out the requested changes

emmatown (commented):
Could you add tests for a few cases:
- Importing a `.d.ts` file with `.d.ts` (as in a `.d.ts` file in `src`)
- Referencing a non-`.d.ts` `.ts` file with `.ts` in a `.d.ts` file
- Using `#imports` in a `.d.ts` file
- Using `#imports` in a `.ts` file with  `onlyEmitUsedTypeScriptDeclarations` but not `importsConditions` (with this new behaviour, I would expect it to be replaced

(I think to make all that work right, some changes will be required, particularly, we need to parse and transform `.d.ts` files)

Andarist:
Sure, I'll get to it - but it might take me a few more days. For now, I need to switch to some other work 😅

Andarist:
@emmatown I addressed half of your list. I still need to figure out how to properly transform declaration files. I tried a bunch of stuff but so far I can't reliably make TS to do anything for `src/**.d.ts` within `program.emit` and that's the only place where I know how to get access to the `TransformationContext` that is required by `visitEachChild`.

I think this PR is in good shape right now and further improvements could be made in a separate PR, WDYT?

emmatown (approved):

PR #558, opened 2023-05-18
Avoid empty reexport statements in `.cjs.mjs` files

Andarist:


packages/cli/src/utils.ts
```diff
@@ -336,6 +336,9 @@ export function dtsTemplate(
 }
 
 function getReexportStatement(namedExports: string[], source: string): string {
+  if (!namedExports.length) {
+    return "";
```
emmatown, line 340:
```suggestion
    return `import ${source};`
```

An entrypoint could (and likely does if it has no exports) have side-effects so we still need to import it.

Andarist: Do u always have to be right? :P pushed out this change with extra tests for this

emmatown (approved):

Issue #259, opened 2020-02-17
Additional entries don't have "types" generated


Andarist:
It's not automatically generated, but even if manually added then it's not validated and cannot be fixed with `preconstruct fix`.

Andarist:
Ok, seems like `.d.ts` is actually generated - but because of the issue mentioned [here](https://github.com/preconstruct/preconstruct/issues/244#issuecomment-587175876) I was thinking that I was looking at another directory (the root entry).

I feel however that maybe it should be validated that "types" got used and throw an error on that (in favor of those type files that are auto-generated). This should ofc happen only for TS projects - adding "types" manually should be supported for packages shipping `.d.ts` without them actually being written in TS.

mitchellhamilton:
> I feel however that maybe it should be validated that "types" got used and throw an error on that (in favor of those type files that are auto-generated). This should ofc happen only for TS projects - adding "types" manually should be supported for packages shipping .d.ts without them actually being written in TS.

Agreed

hasparus:
Hey folks, what's the status of `"types"` field in package.json with Preconstruct? (BTW Preconstruct is _blazing fast_. Wow.)

[TypeScript docs advise to add `"types"` key to package.json](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package). Should `preconstruct fix` add it to point to the generated `.cjs.d.ts` file?

mitchellhamilton:
There's no real need since TypeScript can already resolve it. Is there are a particular reason you want to add it to the package.json?

Andarist:
Even to make it easier for other tools. For example, npm marks packages with TS types with a special badge. Compare:
- https://www.npmjs.com/package/@changesets/git
- https://www.npmjs.com/package/xstate

hasparus:
> Is there are a particular reason you want to add it to the package.json?

What Andarist said, and I vaguely remember PRing/patching niche libraries I wanted to use, because old TS couldn't resolve it.

mitchellhamilton:
Okay, sounds good. Happy to accept a PR adding this(behind a flag for now because it would be a breaking change otherwise)

PR #562, opened 2023-05-29
Fix `types` & `typings` fields

Andarist:
closes https://github.com/preconstruct/preconstruct/issues/259

### DIFF
--- a/.changeset/proud-dodos-carry.md
+++ b/.changeset/proud-dodos-carry.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": minor
+---
+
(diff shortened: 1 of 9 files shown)


emmatown (commented):
I agree with the problem we're solving here but this would break for users who are say writing `.d.ts` files themselves and pointing to `.js` files themselves. I have some ideas on how to solve this problem without causing more problems, I'll come back to this when I have a more concrete plan

Andarist:
Isn't the fix here to just do this when we have TS entrypoints?

emmatown:
yeah, that should work (and it should also be per-entrypoint unlike the other fields)

do you happen to know why both the `typings` and `types` fields exist? I assume some compatibility reason? I'm wondering which one was first? is there a difference in what they do? was there a difference before but there no longer is?

Andarist:
Unfortunately, I don't know why both of those exist. `types` is definitely way more popular.

PR #565, opened 2023-07-02
Use `package.json#exports`

Andarist:


packages/cli/src/cli.ts
```diff
@@ -14,9 +14,7 @@ import {
   BatchError,
 } from "./errors";
 
-// tricking static analysis is fun
-// @ts-ignore
-process["e" + "nv"].NODE_ENV = "production";
+process.env.NODE_ENV = "production";
```
Andarist, line 17: is this even needed now? I'm not sure what this was about

packages/cli/src/rollup-plugins/prod-dev-entry.ts
```diff
@@ -20,10 +20,7 @@ export default function nodeDevProdEntry(): Plugin {
 
         let mainEntrySource = `'use strict';
 
-if (${
-          // tricking static analysis is fun...
-          "process" + ".env.NODE_ENV"
-        } === "production") {
+if (process.env.NODE_ENV === "production") {
```
Andarist, line 23: TODO: recheck if this change is OK when build gets fixed

packages/cli/src/cli.ts
emmatown: It's because some babel plugins might rely on `process.env.NODE_ENV` and Preconstruct is basically always meant for building for "production"

packages/cli/src/rollup-plugins/prod-dev-entry.ts
Andarist: it's OK

.changeset/new-kids-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@preconstruct/cli": minor
+"@preconstruct/eslint-plugin-format-js-tag": minor
+---
+
+`package.json#exports` have been added to limit what
```
Andarist, line 6:
```suggestion
`package.json#exports` have been added to limit what (and how) might be imported from the package. Everything that was meant to be exported for public usage stays exported.
```

.changeset/new-kids-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@preconstruct/cli": minor
+"@preconstruct/eslint-plugin-format-js-tag": minor
+---
+
+`package.json#exports` have been added to limit what (and how) might be imported from the package. Everything that was meant to be exported for public usage stays exported.
```
emmatown, line 6:
```suggestion
`package.json#exports` have been added to limit what (and how) code might be imported from the package.
```

Removed the second thing since I don't really want to emphasise the `package.json#exports`. Really the `package.json#exports` for `@preconstruct/cli` should be `null` or something since it shouldn't really be exported, it should only be accessible via the CLI but it's just easier to not have it like that.

emmatown (approved):
I'm assuming the main reason to do this is to avoid the dev vs prod bundle so it's a bit easier to patch, yeah? it's not like the `package.json#exports` field really matters here?

Andarist:
> I'm assuming the main reason to do this is to avoid the dev vs prod bundle so it's a bit easier to patch, yeah? 

Yeah, exactly.

PR #569, opened 2023-07-04
Add extra runtime checking tests for import conditions

Andarist:
follow up to https://github.com/preconstruct/preconstruct/pull/567


packages/cli/src/build/__tests__/imports-conditions.ts
```diff
@@ -253,6 +260,139 @@ test("imports conditions", async () => {
...
+      "runtime-check.cjs": js`
+        require("@scope/pkg");
+      `,
+    });
+    await build(dir);
+
+    resolve!(dir);
+  });
+
+  describe.each(["cjs", "mjs"] as const)("with %s", (moduleType) => {
```
Andarist, line 357: usually, I hate nested describe blocks a lot but I think it makes sense here when using `.each`

emmatown (approved):

PR #585, opened 2024-01-09
Use narrowed down variable within if's body

Andarist:


packages/cli/test-utils/index.ts
```diff
@@ -352,7 +352,7 @@ export async function testdir(dir: Fixture) {
       const output = dir[filename];
       const fullPath = path.join(temp, filename);
       if (typeof output === "string") {
-        await fs.outputFile(fullPath, dir[filename]);
+        await fs.outputFile(fullPath, output);
```
Andarist, line 355: This is the same value, it just uses the narrowed-down variable. After using this code in another project I noticed that this line errored with certain TS settings/TS versions.

emmatown (approved):

Issue #513, opened 2023-01-25
Question around ESM support for bundlers + node


itsdouges:
Hey folks! Love the library, definitely helps making package interop in monorepos easier. With ESM support what's the current support like?

Looking at commonjs - it uses the babel hook to transpile on the fly so local development works without any extra work. It doesn't look like the ESM output has the same treatment though. Should it? E.g. if I have my package with `"type": "module"`, and have it flow through into the esm module, it won't work during local dev since it's TypeScript and most bundlers (and ts-node) won't transpile node modules without extra effort.

If anyone has insight I'd appreciate it, cheers

jrolfs:
I'm definitely curious to hear @emmatown's perspective on this, but I went somewhat deep into it by way of evaluating [**unbuild**](https://unjs.io/unbuild) as a potential replacement for Preconstruct because it _does_ provide ESM stubs (via [**jiti**](https://unjs.io/jiti)). Unfortunately, those stubs only work for Node.js, and therefore do not address linking packages to be consumed by something that uses a bundler, like Next.js. I'm honestly not even sure if hooking into the experimental and very unstable [loaders API](https://github.com/nodejs/loaders) would work for that case.

You can see a bit more of my related ramblings in a comment [over there](https://github.com/unjs/unbuild/issues/35#issuecomment-1423381785).

emmatown:
To start, `"type": "module"`/`.mjs`/etc. isn't currently supported by Preconstruct. I have thoughts on how it should be supported, specifically it will create an ESM-only package but that's not something I'm aiming to do very soon since the only thing it really allows you to do is use Node ESM-only packages.

Also, Preconstruct is unlikely to ever emit ESM targeting Node + CommonJS like some other tools do because ESM targeting Node (the `import` condition with `"type": "module"`/`.mjs`) + CommonJS doesn't provide value over ESM targeting bundlers (the `module` condition with ESM syntax in a `.js` file without `"type": "module"`) + CommonJS for Node and it adds the problem of the [dual package hazard](https://nodejs.org/dist/latest-v18.x/docs/api/packages.html#dual-package-hazard).

For bundlers, you have to configure them to compile all TS and etc. outside of `node_modules` (even if it's potentially symlinked into `node_modules`) that's e.g. what `@preconstruct/next` does for Next. (this has been true forever with the `dev` command)

[...]

PR #586, opened 2024-01-12
Add `typeModule` experimental flag

emmatown:
For now the output used in `preconstruct dev` will only work directly in Node if you're using no non-standard syntax (and following Node's stricter module resolution) or you're using a bundler that will handle it, I'd like to add a loader and fix that in the future.

Related: #471, #513


site/docs/configuration.md
```diff
@@ -291,6 +291,45 @@ Currently, this defaults to `"namespace"`, this will change to `"default"` in th
...
 
+### `type` (experimental) {#type}
+
+```ts
+"module" | "commonjs";
+```
+
+> If you're just thinking "I want to write code with the native ECMAScript import and export syntax and have my code work for most people", you likely should not use this feature. This feature allows you to build packages that import Node.js ESM only dependencies and therefore only support being imported from ESM in newer versions of Node.js and some bundlers.
+
+The `type` config being set to `"module"` allows you to generate ECMAScript modules(ESM) compatible with Node.js' implementation of ESM (Node.js ESM), most modern bundlers, etc.
```
Andarist, line 302: shouldnt this mention `typeModule` and not `type` with a `"module"` value?

emmatown: `"type": "module"` has to be set as well, the `typeModule` experimental flag is mentioned further down.

packages/cli/src/__tests__/dev.ts
```diff
@@ -687,3 +687,156 @@ test("dev command entrypoint", async () => {
...
+    "src/multiply.ts": js`
+      export let b = "b";
+    `,
+  });
+
+  await dev(dir);
+
+  expect(await getFiles(dir, ["**/dist/**/*"])).toMatchInlineSnapshot(`
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/multiple-entrypoints-multiply.d.ts ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    export * from "../src/multiply";
```
Andarist, line 788: @emmatown I think this should use an explicit extension

PR #590, opened 2024-03-26
Add `use client` directives to the created mjs wrappers

### LINKED ISSUE #182: "use client" not appearing in the ESM entry
- `react-error-boundary` version: 4.0.13 (latest)
- `node` version: v20
- `npm` version: v10

Relevant code or config

```js
import { ErrorBoundary } from 'react-error-boundary'
```

What you did:

Using the node loader from react-server-dom-esm (https://github.com/facebook/react/blob/e373190faf3b994707f09488c1a7832f4a91e15a/packages/react-server-dom-esm/src/ReactFlightESMNodeLoader.js#L471-L481)

What happened:

`react-error-boundary` is not treated as a client component because the "use client" pragma doesn't appear in [the ESM module](https://unpkg.com/browse/react-error-boundary@4.0.13/dist/react-error-boundary.cjs.mjs) (only [the CJS module](https://unpkg.com/browse/react-error-boundary@4.0.13/dist/react-error-boundary.cjs.js)).

Problem description:

The problem is that node loaders can only transform native ESM modules. So the "use client" pragma needs to appear in the ESM module. Something in the build process doesn't make that work though.

Suggested solution:

My workaround is to create a separate module in my own codebase that adds the "use client" pragma and import that instead. I would love to not have to do that.

Andarist:
fixes https://github.com/bvaughn/react-error-boundary/issues/182

I'm not 100% sure if this should be done at all though. It doesn't solve the case when the directive is in non-entrypoint file.

### DIFF
--- a/packages/cli/src/build/__tests__/other.ts
+++ b/packages/cli/src/build/__tests__/other.ts
@@ -1048,6 +1048,134 @@ test("use client", async () => {
   `);
 });
 
+test("use client (importConditionDefaultExport: default)", async () => {
+  const dir = await testdir({
+    "package.json": JSON.stringify({
+      name: "pkg",
+      main: "dist/pkg.cjs.js",
+      module: "dist/pkg.esm.js",
+      exports: {
+        ".": {
+          types: {
+            import: "./dist/pkg.cjs.mjs",
+            default: "./dist/pkg.cjs.js",
+          },
+          module: "./dist/pkg.esm.js",
+          import: "./dist/pkg.cjs.mjs",
+          default: "./dist/pkg.cjs.js",
+        },
+        "./package.json": "./package.json",
+      },
+      preconstruct: {
+        exports: {
+          importConditionDefaultExport: "default",
+        },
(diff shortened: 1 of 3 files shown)


packages/cli/src/build/__tests__/other.ts
```diff
@@ -1090,6 +1218,65 @@ test("use client as entrypoint", async () => {
...
+    'use strict';
+
+    Object.defineProperty(exports, '__esModule', { value: true });
+
+    const a = true;
+
+    exports.a = a;
+
+    ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ dist/pkg.cjs.mjs ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
+    'use client';
```
Andarist, line 1266: now the use client directive is in this file and the one that it reexports (the CJS file) - I think it should be harmless, right?

PR #599, opened 2024-07-17
Remove dependency on meow

### LINKED ISSUE #598: Remove `meow` package in favor of parseArgs from node:util
Hello,

Looking at the way `@preconstruct/cli` uses `meow`, it should be easy to replace it with the native [parseArgs](https://nodejs.org/docs/latest-v20.x/api/util.html#utilparseargsconfig) that is available since node v16.17.0 and v18.3.0 (although it's marked experimental there, it seems to work without issue. It has no experimental flag since version v20.0.0).

Given that there is no args parsing in `@preconstruct/cli`, it's also possible to directly use `process.argv`. 

Removing `meow` would remove 62 dependencies from the installation of `@preconstruct/cli`: https://npmgraph.js.org/?q=meow%407 - this is a whole 30% of all dependencies installed by `@preconstruct/cli`.

If there is interest in this from the maintainers, I can provide a PR. If not, please feel free to close this issue.

Cheers!

VanTanev:
closes #598

### DIFF
--- a/.changeset/dry-ladybugs-shake.md
+++ b/.changeset/dry-ladybugs-shake.md
@@ -1,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Remove unnecessary dependency on `meow` CLI parsing library, as @preconstruct/cli only takes a single positional argument denoting a preconstruct command.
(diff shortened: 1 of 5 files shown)


VanTanev:
I think that the failing CI is unrelated to the changes in the PR

emmatown (approved):
Thanks!

PR #604, opened 2024-10-10
Fix outputting `.d.ts` files with extension-less imports that are invalid when using the experimental flag `typeModule`

emmatown:
cc @Andarist re https://github.com/preconstruct/preconstruct/pull/586#discussion_r1788573858

I'm surprised I didn't hit this earlier tbh, particularly the build case of it since I've been using this on a project for months (makes sense though since I basically always use bundler over node16)

Also, just a random note: I know there's some weird stuff here where in some places we handle `cjs`/`mjs`/the ts variants and elsewhere we don't, I believe those extensions just don't work in preconstruct right now. I might look into that in the future though not very high priority since we wouldn't allow mixing module types so there isn't really a point in using them.

### DIFF
--- a/packages/cli/src/dev.ts
+++ b/packages/cli/src/dev.ts
@@ -82,6 +82,10 @@ export async function entrypointHasDefaultExport(
   return false;
 }
 
+function replaceTsExtensionWithJs(pathname: string) {
+  return pathname.replace(/\.(?:tsx|([mc])?ts)$/, ".$1js");
+}
+
 export async function writeDevTSFiles(
   entrypoint: Entrypoint,
   hasDefaultExport: boolean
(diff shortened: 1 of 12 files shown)


packages/cli/src/dev.ts
```diff
@@ -82,6 +82,10 @@ export async function entrypointHasDefaultExport(
   return false;
 }
 
+function replaceTsExtensionWithJs(pathname: string) {
+  return pathname.replace(/\.(?:tsx|([mc])?ts)$/, ".$1js");
```
Andarist, line 86: nit: as far as I can see `.cts` and `.mts` cases are not covered with tests ;p

Andarist:
Thanks! ❤️

Andarist:
> (makes sense though since I basically always use bundler over node16)

I'm not entirely sure about this but it feels like a bundler might be a bad choice for monorepos that are meant to publish packages. If you depend on anything external you can't trust the bundler resolution - those imports are left as external in the output and we want to make sure that they will continue to work for our consumers

packages/cli/src/dev.ts
emmatown: Yeah, mentioned those extensions in the PR description, I believe they just don't work right now and fixing that isn't really a high priority tbh since they aren't really useful in Preconstruct's context and we'd want to ban mixing module types anyway

PR #607, opened 2024-10-13
[dependencies-replacement] Replace fs-extra with node:fs/promises

VanTanev:
References #608

Removes 5 dependencies


VanTanev:
I need some guidance if the changes being introduced by replacing dependencies should be considered a minor change or a major one.

It seems that previously, when `fast-glob` was introduced, it was treated as a major change.

PR #615, opened 2025-02-24
Enforce that the main/module/browser/umd:main fields are not present when using the "type": "module" mode

emmatown:
Ref https://github.com/changesets/changesets/pull/1482#discussion_r1966853781

### DIFF
--- a/packages/cli/src/fix.ts
+++ b/packages/cli/src/fix.ts
@@ -44,7 +44,13 @@ async function fixPackage(pkg: Package): Promise<() => Promise<boolean>> {
     }
   }
 
-  if (!pkg.isTypeModule()) {
+  if (pkg.isTypeModule()) {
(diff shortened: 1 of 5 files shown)


packages/cli/src/fix.ts
```diff
@@ -44,7 +44,13 @@ async function fixPackage(pkg: Package): Promise<() => Promise<boolean>> {
     }
   }
 
-  if (!pkg.isTypeModule()) {
+  if (pkg.isTypeModule()) {
+    keys(fields).forEach((field) => {
+      if (fields[field]) {
+        delete (pkg.json as any)[field];
```
emmatown, line 50: Note this is done on `pkg.json` instead of the entrypoints since the entrypoint `package.json`s(besides the package's one itself) are removed automatically already elsewhere

Andarist (approved):

PR #626, opened 2026-05-08
Fix UMD resolution for mixed inline and `@babel/runtime` Babel helpers

Andarist:


.changeset/forty-deers-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Fix UMD resolution for mixed inline and @babel/runtime Babel helpers
```
Andarist:
```suggestion
Fix UMD resolution for mixed inline and `@babel/runtime` Babel helpers
```

.changeset/forty-deers-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preconstruct/cli": patch
+---
+
+Fix UMD resolution for mixed inline and `@babel/runtime` Babel helpers
```
emmatown, line 5: I don't get the mixed part of this, isn't this just fixing the case of using `@babel/runtime` for UMD builds and the inline case already worked?

